PR #17, opened 2018-09-26
Field feedback

taimur38:
Incorporating usage feedback that we got from our field visit on 9/25. please check on https://mis.metal.fish and let me know if I missed anything

### DIFF
--- a/elixir/lib/mix/tasks/migrate.ex
+++ b/elixir/lib/mix/tasks/migrate.ex
@@ -1,0 +1,41 @@
+defmodule Mix.Tasks.Migrate do
+	use Mix.Task
+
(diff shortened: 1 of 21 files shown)


Ayesha19:
Teacher's profile:
-[ ] Place qualifications before experience
image
-[ ] The 'add new teacher' remains filled with the last added teacher's data.

Ayesha19:
Subjects:
- [ ] The name of subject should be a compulsory field. so that none can save without name.
image

Class:
- [ ] I kept the class year of play group '0' but it is coming in between the list.
image

Attendance:
- [ ] why the date has two dropdown and a close cross?
image

PR #22, opened 2018-11-08
Added Search functionality

taimur38:


web/src/components/List/index.js
```diff
@@ -1,22 +1,66 @@
...
+		this.state = {
+			items: [],
+		}
+	}
+	
+	filterlist = (e) => { 
+
+		const updatedList = this.props.children.filter((item) => {
+			//console.log(item);
+			return item.props.children.toLowerCase().search(e.target.value.toLowerCase())!== -1;
```
taimur38: Nothing guarantees that `item.props.children` will be a string. It would break if I passed in children which had several children of their own, or had no children. So this way won't work. look at the Dropdown component for inspiration on how you should restructure this.

web/src/components/List/index.js
```diff
@@ -1,22 +1,66 @@
...
+		});
+		//console.log(updatedList)
+		this.setState({items : updatedList});
+	}
+
+	componentDidMount(){
+		let items = this.props.children; 
+		this.setState({ items });
+		//console.log(items);
```
taimur38: This is a bad pattern - you should never have to convert props into state.

taimur38: A simpler way to do this, is to keep the value of the input as state, and do your filter in the render function

web/src/components/List/index.js
```diff
@@ -7,37 +7,33 @@ import './style.css'
...
-	filterlist = (e) => { 
-
-		const updatedList = this.props.children.filter((item) => {
-			//console.log(item);
-			return item.props.children.toLowerCase().search(e.target.value.toLowerCase())!== -1;
-		});
-		//console.log(updatedList)
-		this.setState({items : updatedList});
+	getLabel = (item) => {
+		return item.props.children;
```
taimur38: Same comment from before still applies.

> Nothing guarantees that item.props.children will be a string. It would break if I passed in children which had several children of their own, or had no children. So this way won't work. look at the Dropdown component for inspiration on how you should restructure this. 

The parent of the `List` Component is passing in the children. There is no restriction on the types of children that can be passed in. So the `List` Component has no way of knowing how to get the label out of the children. The parent of `List` should be telling it how via props.

web/src/components/List/index.js
```diff
@@ -7,37 +7,33 @@ import './style.css'
...
+		this.setState({filterText:e.target.value});
 	}
 
 	render() {
-		return <div className="list-wrap">
-			<input type="text" placeholder="Search" onChange={this.filterlist}/>
 
+		const filteredList = this.props.children.filter(item => {
+			return this.getLabel(item).toLowerCase().includes(this.state.filterText.toLowerCase());
+		})
```
taimur38: Much better way of implementing than before, where you had kept the filtered list as state. Here the filter is **derived** at `render` from the state of "whatever is in the text box" (`filterText`). The way you were doing it before relied on hiding the fact that the input component had its own internal state, and copying props to state.

web/src/components/List/index.js
```diff
@@ -7,37 +7,33 @@ import './style.css'
...
+		
+		return <div className="list-wrap">
+			<input type="text" placeholder="Search" onChange={this.onChange}/>
+			
 			{ this.props.create ? <Create to={this.props.create} text={this.props.createText} /> : false }
 			
 			<div className="list">
 				{
-					this.state.items.map(C => <div className="list-row" key={Math.random()}>{C}</div>)
+					filteredList.map(C => <div className="list-row" key={Math.random()}>{C}</div>)
```
taimur38: I used `Math.random()` as a hack, but the better way to do this is shown in `components/Dropdown`. See if you can research why that is the case, i.e what does reacts "key" do, why is it required?

PR #25, opened 2018-11-14
Redesign backend

taimur38:
Create a new table for writes so we have a log of each write per school in the db. Don't send db snapshots - only send the new writes back and forth.

Needs more testing


taimur38:
-[ ] test one very out of date client for latest feature
-[ ] check memory limits and garbage collection behaviour
-[ ] test clients that are on different version - can we roll this backend out simultaneously with old client and support both.
-[ ] can a client that was on an old version transition into this without weird behavior. what happens when multiple clients switch at staggered times.
-[ ] all other tests related to difficulty of merging this in later.

PR #33, opened 2018-12-03
section by default, default class names

taimur38:


web/src/utils/getSectionsFromClasses.js
```diff
@@ -5,13 +5,13 @@ export const getSectionsFromClasses = (classes) => {
 		.reduce((agg, c) => {
 			// each section
 			return [...agg, ...Object.entries(c.sections)
-				.reduce((agg2, [id, section]) => {
+				.reduce((agg2, [id, section], i, arr) => {
```
raoaliahmad: You're changing getSectionsFromClasses, there are some other classes that are using this. Are they working fine?

taimur38: yup

PR #48, opened 2018-12-14
removed the print error and removed the print preview

raoaliahmad:


web/src/modules/Class/Single/Reports/index.js
```diff
@@ -1,17 +1,12 @@
...
 import './style.css'
 
-const ClassReports = ({ match, classes, students, exams, settings, sms_templates }) => {
-
-	const id = match.params.id;
-	const start = parseFloat(match.params.start)
-	const end = parseFloat(match.params.end)
-
+export const ClassReports = (props) => {
+	const {id, classes, students, exams, settings, sms_templates, start, end} = props;
```
taimur38: this, why change

taimur38: as in, you should destructure in the function signature

web/src/index.js
```diff
@@ -11,7 +11,6 @@ import Routes from './routes'
 import { saveDB, loadDB } from 'utils/localStorage';
 import Syncr from 'syncr'
 
-const debug_host = 'wss://ab9eae32.ngrok.io'
```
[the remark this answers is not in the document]
raoaliahmad: so you dont get conflicts while merging

web/src/modules/Class/Single/Reports/index.js
raoaliahmad: Yeah I  was checking something out so I did it :D

web/src/utils/Hyphenator.js
```diff
@@ -0,0 +1,15 @@
...
+		if(str.substring(5,6) !== "" && str.substring(5,6) !== "-")
+			str = str.substr(0, 5) + "-" + str.substr(5);
+		
+		if(str.substring(13,14) !== "" && str.substring(13,14) !== "-")
+			str = str.substr(0, 13) + "-" + str.substr(13);
+
+		return str;
+	}
+}
+export default Hyphenator;
```
taimur38: Look into the difference between `export const Hyphenator` and `export default`. Why have both? which one is appropriate?

web/src/modules/Teacher/Single/Create/index.js
```diff
@@ -150,16 +154,12 @@ class CreateTeacher extends Component {
 		})
 	}
 
-/* 	formatCNIC = () => {
-
-		// substring
-		console.log('substr(4,6): ',this.state.profile.CNIC.substr(5,6))
+	addHyphens = (path) => () => {
+		
+		let str = Dynamic.get(this.state, path);
```
taimur38: What is difference between let and const? which one is appropriate here

web/src/modules/Class/Single/ReportsMenu/index.js
```diff
@@ -34,10 +34,26 @@ class ClassReportMenu extends Component {
 					<label>End Date</label>
 					<input type="date" onChange={this.report_former.handle(["end"])} value={moment(this.state.report_dates.end).format("YYYY-MM-DD")} placeholder="End Date" />
 				</div>
-				<Link className="button grey block" to={`reports/${moment(this.state.report_dates.start).unix() * 1000}/${moment(this.state.report_dates.end).unix() * 1000}`}>Preview</Link>
```
taimur38: The Link has been removed. but how did the link work before? what would happen if you went to this url still?

raoaliahmad: 'Link' creates a link to the route you have defined.

PR #52, opened 2018-12-19
Added "Delete Button" in Class

raoaliahmad:


web/src/actions/index.js
```diff
@@ -105,6 +105,14 @@ export const createEditClass = newClass => dispatch => {
...
 }
 
+export const deleteClass = (Class) => dispatch => {
+	dispatch(createDeletes([
+		{
+			path:["db", "classes", Class.id]
+		}
+	]))
+}
+
```
taimur38: modify this method so that the logic of setting student section_id = "" happens here, and happens as a single `dispatch(createMerges(....))`. then, dispatch the createDelete

web/src/modules/Class/Single/Create/index.js
```diff
@@ -165,8 +164,34 @@ class SingleClass extends Component {
...
+		Object.values(this.props.students)
+			.forEach(student => Object.keys(Class.sections)
+					.forEach(section => 
+						{ 
+							if(section === student.section_id) 
+							this.addStudent("")(this.props.students[student.id])
+						})
+					)
+
+		this.props.removeClass(Class)
```
taimur38:
```javascript

Object.values(this.props.students)
    .forEach(student => Class.sections[student.section_id] != undefined ? this.addStudent("")(student)
```

web/src/actions/index.js
taimur38:
```javascript
export const deleteClass = (Class) => (dispatch, getState) => {
    const state = getState();

   const students = Object.values(state.db.students).filter(x => Class.sections[x.section_id].......
```

PR #55, opened 2018-12-24
Grades with marks

raoaliahmad:


web/src/actions/index.js
```diff
@@ -223,8 +223,11 @@ export const mergeExam = (exam, class_id, section_id) => dispatch => {
 		.reduce((agg, [student_id, score]) => ([
 			...agg,
 			{
-				path: ["db", "students", student_id, "exams", id, "score"],
-				value: score
+				path: ["db", "students", student_id, "exams", id ],
+				value: {
+					score: score.marks, 
+					grade: score.grade
+				}
```
taimur38: rename the `score` variable here to something like `student_mark`. then it should be `student_mark.score` and `student_mark.grade`

web/src/modules/Student/Single/Marks/index.js
```diff
@@ -119,6 +120,7 @@ export const StudentMarks = ({student, exams, settings, startDate=0, endDate=mom
 						<div>{student.exams[exam.id].score}</div>
 						<div >{exam.total_score}</div>
 						<div>{(student.exams[exam.id].score / exam.total_score * 100).toFixed(2)}</div>
```
taimur38: are we calling it score or marks??

web/src/actions/index.js
taimur38: in other places we use `marks` and `grade`. here we are using `score` and `grade`. we should use only `score` because it already exists, and we shouldn't be using the word `marks` to describe the numerical score of the test in other sections.

PR #56, opened 2018-12-27
Permission system

raoaliahmad:


web/src/modules/Settings/index.js
```diff
@@ -11,7 +13,24 @@ const defaultSettings = {
...
+		student_attendance_module:	{ teacher: true  }, //added
+		teacher_attendance_module:	{ teacher: true  }, //didn't needed to add as only admin can see it
+		sms_module:					{ teacher: true  }, //added
+		marks_module:				{ teacher: true  }, //added
+
+		student_profile:			{ teacher: true  }, 
+		teacher_profile:			{ teacher: false  },
+		class_profile:				{ teacher: true  },
+		student_fee_information:	{ teacher: false }
+	}
```
taimur38: 1. these are ambiguous names. there are 4 types of operations - Create, Read, Update, Delete. does true/false mean they can do all of those things? what is the difference between fee and student_fee_information? 

2. shouldn't use both underscores and camelCase - should pick one. Here we've used camelCase for all the other fields, so you should use it as well.

web/src/modules/Settings/index.js
```diff
@@ -131,10 +222,12 @@ class Settings extends Component {$Name
...
-		settings: state.db.settings, 
+		settings: state.db.settings,
+		user: state.db.faculty[state.auth.faculty_id], 
 		sms_templates: state.db.sms_templates 
 	}), 
 	dispatch => ({
 		saveTemplates: templates => dispatch(createTemplateMerges(templates)),
-		saveSettings: settings => dispatch(mergeSettings(settings))
+		saveSettings: settings => dispatch(mergeSettings(settings)),
+		savePermissions: permissions => dispatch(mergePermissions(permissions))
```
taimur38: should not need separate savePermissions action

web/src/modules/Student/List/index.js
```diff
@@ -72,15 +83,19 @@ export const StudentList = ({ classes, students, settings, forwardTo, history })
 			Component = {StudentItem}
 			create = {create} 
 			createText = {"Add new Student"} 
-			toLabel = {toLabel} /> 
+			toLabel = {toLabel}
+			create_permission = {admin ? true : permissions.addStudent.teacher}
```
taimur38: should not have extra prop for List

web/src/modules/Student/List/index.js
```diff
@@ -72,15 +83,19 @@ export const StudentList = ({ classes, students, settings, forwardTo, history })
...
 		<div className="print button" onClick={() => window.print()}>Print</div>
 	</div>
 }
 
 export default connect((state, { location }) => ({ 
 	students: state.db.students,
+	admin: state.db.faculty[state.auth.faculty_id].Admin,
 	classes: state.db.classes,
 	settings: state.db.settings,
+	permissions: state.db.settings.permissions,
```
taimur38: this is already nested under settings, don't need to add it explicitly

web/src/modules/Student/Single/index.js
```diff
@@ -24,7 +25,7 @@ export default class StudentPage extends Component {
 				{ loc === "new" ? false : 
 				<div className="row tabs">
 					<Link className={`button ${loc === "profile" ? "red" : false}`} to="profile" replace={true}>Profile</Link>
-					<Link className={`button ${loc === "payment" ? "green": false}`} to="payment" replace={true}>Payment</Link>
+					{this.props.user.Admin ? <Link className={`button ${loc === "payment" ? "green": false}`} to="payment" replace={true}>Payment</Link> : false}
```
taimur38: made all those permissions but in the end, only checking if they are an admin here?

PR #69, opened 2019-01-14
Added auto promotion

raoaliahmad:
1) Auto Promote Completed, 
2) added checks in result for prev class marks
3) resolved sms not picking single student result


.vscode/settings.json
```diff
@@ -15,5 +15,6 @@
 		"**/bower_components": true,
 		"**/_build": true,
 		"**/deps": true
-	}
+	},
+	"editor.fontSize": 17
```
taimur38: Should keep this in your personal settings, not on project settings

web/src/actions/index.js
```diff
@@ -87,54 +89,91 @@ export const createSchoolLogin = (school_id, password) => (dispatch, getState, s
...
+
+	const studentMerges = students.map(s => (
+		{ path: ["db", "students", s.id, "section_id"], value: newSectionId }
+	))
+	const historyMerges = history.map(s => (
+		{
+			path: ["db", "students", s.id, "classHistory"],
+			value: {
+				...s.classHistory,
+				[prevClassId]: {
```
taimur38: should this be previous class id or section id?

web/src/actions/index.js
```diff
@@ -87,54 +89,91 @@ export const createSchoolLogin = (school_id, password) => (dispatch, getState, s
...
+					"class_id": prevClassId,
+					"section_id": prevSectionId,
+					"className": prevClass.name,
+					"promotionDate": time,
```
taimur38: we have been using section_id and class_id -- have to pick either underscore or camel case convention. we use underscores in the db

web/src/modules/Class/Single/ReportsMenu/index.js
```diff
@@ -25,27 +25,26 @@ class ClassReportMenu extends Component {
...
 
-		for(let s of relevant_students)
-		{
-			for(let e of Object.values(this.props.exams))
-			{ 
-				if(e.section_id === s.section_id){
+		for (let s of relevant_students) {
+			for (let e of Object.values(this.props.exams)) {
+				
+				if (s.exams !== undefined && s.classHistory[e.class_id] === undefined && s.exams[e.id] !== undefined) {
```
taimur38: is this supposed to be s.classHistory[e.class_id] !== undefined? why does it need to be in class history

web/src/modules/Marks/SingleExam/index.js
```diff
@@ -64,7 +64,7 @@ class SingleExam extends Component {
 				const exam = student.exams[exam_id]  
 				return {
 					...agg,
-					[id]: exam ? exam.score : ""
+					[id]: exam || { score:"", grade: "" }
```
taimur38: out of date

web/src/modules/SMS/SmsOptions/ToSingleClass.js
```diff
@@ -8,18 +8,18 @@ import former from 'utils/former'
...
-		  text: ""
-	  }
+	this.state = {
+		selected_section_id: "",
+		selected_student_number: "",
+		text: ""
+	}
 
-	  this.former = new former(this, [])
+	this.former = new former(this, [])
```
taimur38: why are so many random lines of code changed just by spacing. this makes it really hard to read through the actual changes

web/src/modules/SMS/SmsOptions/ToSingleClass.js
```diff
@@ -8,18 +8,18 @@ import former from 'utils/former'
...
 	}
 
-  	render() {
+render() {
```
taimur38: this should not be happening

web/src/modules/Promotion/index.js
```diff
@@ -0,0 +1,161 @@
...
+                filterOut: arr
+            })
+        }
+        else {
+            const arr = this.state.filterOut.filter(s_id => s_id !== e.target.value);
+            this.setState({
+                filterOut: arr
+            })
+        }
+    }
```
taimur38: why is there an oncheck and former in the same place.

also looks like filterOut should be an object or set, and not an array.

web/src/modules/Promotion/index.js
```diff
@@ -0,0 +1,161 @@
...
+                    text: "Please Select a Higher order class to Promote"
+                }
+            }), setTimeout(() => {
+                this.setState({
+                    banner: {
+                        active: false
+                    }
+                })
+            }, 1000)
+        }
```
taimur38: why do we care if the promote up or down

web/src/modules/Promotion/index.js
```diff
@@ -0,0 +1,161 @@
...
+        this.props.promote(sections[this.state.promoteFrom].class_id, this.state.promoteFrom, this.state.promoteTo, this.state.filterOut)
```
taimur38: overall structure: why say that we are promoting this class section, and also specifying the promote from and the promote to? 

why not just have a list of students who will be sent to new classes?

web/src/utils/localStorage.js
```diff
@@ -9,17 +9,17 @@ const defaultTemplates = () => ({
...
 	lastSnapshot: 0,
 	db: {
-		faculty: { },
-		users: { }, // username: passwordhash, permissions, etc.  
-		students: { },
-		classes: { }, // id: { name, class, teacher_id, subjects: { name: 1 } },
+		faculty: {},
+		users: {}, // username: passwordhash, permissions, etc.  
+		students: {},
+		classes: {}, // id: { name, class, teacher_id, subjects: { name: 1 } },
```
taimur38: why

web/src/utils/localStorage.js
```diff
@@ -156,24 +156,48 @@ const addFacultyID = state => {
...
+
+	for (let s_id of student_ids) {
+		state.db.students[s_id] = {
+			...state.db.students[s_id],
+			classHistory: {}
+		}
+	}
+	return state
+}
+
```
taimur38: we learned that we need to be more careful with this. have you actually foolproofed this

.vscode/settings.json
raoaliahmad: I think this got pushed accidentaly

web/src/actions/index.js
raoaliahmad: prev class Id because they are being promoted to a new class and section id will be saved inside the object

web/src/actions/index.js
raoaliahmad: Yeah I should have been more careful with naming

web/src/modules/Class/Single/ReportsMenu/index.js
raoaliahmad: this line is checking if the exam is not from the previous class because an exam has the class id, so if the exams class id is same as the class id of the students class history that means these are the prev class exams.

web/src/modules/SMS/SmsOptions/ToSingleClass.js
raoaliahmad: Yeah I was just trying to make the indentation right so I did format document in vs code which formatted it but I totally forgot about how it would affect the readability.

web/src/modules/SMS/SmsOptions/ToSingleClass.js
raoaliahmad: Again formatting Issue

web/src/modules/Promotion/index.js
raoaliahmad: I added `onCheck` because I had to push data in an array and I don't think that's possible using former

web/src/modules/Promotion/index.js
raoaliahmad: I talked to ayesha and she didn't wanted to take any chances and Kareem also told me that right now just do promote and why would they even promote a student that they would want to demote them

web/src/modules/Promotion/index.js
taimur38: Nothing wrong with section-wise promoting on the front-end. but in code its not necessary - it should just be dealing with lists of students, and that list can be filtered by the component

PR #70, opened 2019-01-17
Made the Pull request :D

raoaliahmad:
- report card filter
- remarks in result card
- fix for saving marks 
- analytics endpoints


web/src/modules/Marks/SingleExam/index.js
```diff
@@ -64,7 +64,7 @@ class SingleExam extends Component {
 				const exam = student.exams[exam_id]  
 				return {
 					...agg,
-					[id]: exam ? exam.score : ""
+					[id]: exam || { score: "", grade: "" }
```
taimur38: Need to add remarks field here also

PR #79, opened 2019-01-24
Added Instant fee editing

raoaliahmad:
Sorry for the bad branch name

### DIFF
--- a/web/src/components/List/index.js
+++ b/web/src/components/List/index.js
@@ -19,9 +19,11 @@ class List extends Component {
 	render() {
 		const {items, toLabel, Component } = this.props;
 		
-		const filteredList = items.filter(item => {
-			return toLabel(item).toLowerCase().includes(this.state.filterText.toLowerCase());
-		}).sort((a,b) => toLabel(b).localeCompare(this.state.filterText) - toLabel(a).localeCompare(this.state.filterText))
+		const filteredList = items
+			.filter(item => {
+				return toLabel(item).toLowerCase().includes(this.state.filterText.toLowerCase());
+			})
+			.sort((a,b) => toLabel(b).localeCompare(this.state.filterText) - toLabel(a).localeCompare(this.state.filterText))
 
 		return <div className="list-wrap">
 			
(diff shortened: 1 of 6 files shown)


web/src/components/List/index.js
```diff
@@ -19,9 +19,11 @@ class List extends Component {
 	render() {
 		const {items, toLabel, Component } = this.props;
 		
-		const filteredList = items.filter(item => {
-			return toLabel(item).toLowerCase().includes(this.state.filterText.toLowerCase());
-		}).sort((a,b) => toLabel(b).localeCompare(this.state.filterText) - toLabel(a).localeCompare(this.state.filterText))
+		const filteredList = items
+			.filter(item => {
+				return toLabel(item).toLowerCase().includes(this.state.filterText.toLowerCase());
```
taimur38: Can we add a `toLabel(item) && ` before to short circuit any possible undefined error

web/src/modules/Student/Single/Create/index.js
```diff
@@ -165,9 +165,39 @@ class SingleStudent extends Component {
...
+			if( curr_payment_date === curr_month && 
+				corresponding_fees === undefined) 
+			{
+				const {[p_id]:removed, ...nextPayment} = student.payments
+				student.payments = nextPayment;
+			}
+			else if(curr_payment_date === curr_month && 
+				current_payment.type === "OWED" && 
+				corresponding_fees.period === "MONTHLY" &&
+				Math.abs(current_payment.amount) !== Math.abs(corresponding_fees.amount) )
```
taimur38: Do you need a parsefloat on the fee amount before abs? Think it is saved as a string

web/src/modules/Student/Single/Create/index.js
```diff
@@ -165,9 +165,39 @@ class SingleStudent extends Component {
...
+			{
+				student.payments[p_id] = {
+					amount: corresponding_fees.type !== "SCHOLARSHIP" ?  corresponding_fees.amount : (-1 * corresponding_fees.amount), // check if scholarship, then make negative (just like checkstudentdues function)
```
taimur38: Also here can we make sure fee amount is a number and not string

PR #84, opened 2019-01-30
Added Client id in local Storage

raoaliahmad:


web/src/utils/localStorage.js
```diff
@@ -37,17 +38,21 @@ const initState = {
...
 	try {
-		const serialized = localStorage.getItem('db');
+		const serialized = localStorage.getItem('db');		
 		if (serialized === null) {
 			console.log('null')
 			return initState;
 		}
-		
+
+		const client_id = localStorage.getItem('client_id')
```
taimur38: if this is undefined, we should either generate a new one or check if there was a client_id already in the old state

PR #87, opened 2019-02-05
Added SMS History

raoaliahmad:


web/src/modules/Class/Single/ReportsMenu/index.js
```diff
@@ -24,21 +26,58 @@ class ClassReportMenu extends Component {
...
+		})
+
 		const subjects = new Set()
-		const exams = new Set()
+		const examSet = new Set()
 		
-		const relevant_students = Object.values(this.props.students)
-			.filter(s => this.props.curr_class.sections[s.section_id] !== undefined)
+		const relevant_students_set = Object.values(students)
+			.filter(s => curr_class.sections[s.section_id] !== undefined)
```
taimur38: what is the difference between this and the other relevant_students

web/src/modules/SMS/SmsOptions/ToAllStudents.js
```diff
@@ -14,8 +17,22 @@ export default class ToAllStudents extends Component {
...
 	}
 
-	render() {
+	smsHistory = (messages) =>{
+		if(messages.length === 0){
+			console.log("No Message to Log")
+			return
+		}
+		const historyObj = {
+			date: moment.now(),
```
taimur38: same here, in all the historyObj use `date: new Date().getTime()`

web/src/modules/SMS/index.js
```diff
@@ -56,7 +56,7 @@ class SMS extends Component {
 		}
 
 		console.log('send message', text, number);
-		this.props.sendMessage(text, number);
+		this.props.sendMessage(text, number, this.state.smsFilter);
```
taimur38: are you sure that `this.state.smsFilter` is exactly the same type as what you have put for the other method.

PR #91, opened 2019-02-08
Added Tags in students and Higher Class Promotion

raoaliahmad:


web/src/actions/index.js
```diff
@@ -78,12 +78,27 @@ export const promoteStudents = (promotion_map, section_metadata) => dispatch =>
...
 	dispatch(createMerges([
 		...Object.entries(promotion_map)
-			.map(([student_id, { current, next }]) => ({
-				path: ["db", "students", student_id, "section_id"],
-				value: next
-			})),
+			.map(([student_id, { current, next }]) => {	
+				if(next === "FINISHED_SCHOOL"){
+					//to set finishing students status inactive
+					setInactive.push({
```
taimur38: I don't like the idea of the map function having a side effect, in this case filling another array. Even in this function, we call 3 different map functions on the same initial array `...Object.entries(promotion_map)`. I would maintain the pattern and just add a 4th map from the same source array. 

What should really be done though is converting these 4 `map` calls into one `reduce` call.

web/src/actions/index.js
```diff
@@ -368,4 +391,16 @@ export const logSms = (history) => dispatch => {
 			value : history
 		}
 	])) 
+}
+
+export const addTag = (path, tag) => dispatch => {
```
taimur38: this isn't a useful function because it's doing exactly what createMerges does. 

I want to add a tag to a student. So i should just pass in the text of the tag and the student object. from there the action should handle everything

web/src/utils/localStorage.js
```diff
@@ -192,10 +192,25 @@ const checkPermissions = state => {
...
+		if( state.db.students[s_id].tags === undefined )
+		{
+			state.db.students[s_id] = {
+				...state.db.students[s_id],
+				tags: {}
+			}
+		}
+	} 
+	return state
+}
```
taimur38: this isn't necessary... tags are nested enough that we don't need to add a script like this to handle it being undefined

PR #94, opened 2019-02-13
Device name structure updated

raoaliahmad:


web/src/modules/Settings/index.js
```diff
@@ -36,6 +36,10 @@ class Settings extends Component {
 			permissions: {
 				...defaultPermissions,
 				...(props.settings || defaultSettings).permissions
+			},
+			devices:{
+				...defaultSettings.devices,
+				...(props.settings || defaultSettings).devices
```
taimur38: `devices: (props.settings ? props.settings.devices || {})`

PR #100, opened 2019-02-21
Fee editing from ledger

raoaliahmad:


web/src/modules/Student/Single/Fees/index.js
```diff
@@ -146,6 +147,9 @@ class StudentFees extends Component {
 
 
 	}
+	onSave =()=>{
```
taimur38: `onSave = () => {`

web/src/modules/Student/Single/Fees/index.js
```diff
@@ -210,10 +216,34 @@ class StudentFees extends Component {
...
+				</div>
+					{filteredPayments
+						.map(([id, payment]) => {
+							return <div className="payment" key={id}>								
+								<div className="table row">
+									<div>{moment(payment.date).format("DD/MM")}</div>
+									<div>{payment.type === "SUBMITTED" ? "Payed" : payment.type === "FORGIVEN" ? "Need Scholarship" : payment.fee_name || "Fee"}</div>
+									
+									{ moment(payment.date).format("MM/YYYY") === curr_month && payment.type !== "SUBMITTED" ? 
+										<input style={{border: "none"}} type="number" placeholder={payment.type === "OWED" ? `${payment.amount}` : `-${payment.amount}`} {...this.Former.super_handle(["edits", id])}/>
```
taimur38: Don't use the placeholder for the existing value - set up your "edits" object in state in the constructor and use the value here. Also you will need to make it more obvious that this is an editable field

PR #102, opened 2019-02-24
Diary module

raoaliahmad:


web/src/modules/Diary/index.js
```diff
@@ -0,0 +1,173 @@
...
+            <div className="title">Diary</div>
+            <div className="form">
+                <div className="divider">Send Diary for {moment().format("DD/MM/YYYY")}</div>
+                
+                <div className="section">
+                    <div className="row">
+                        <label>Select Class/Section</label>
+                            <select {...this.former.super_handle(["selected_section_id"])}>
+                                {
+                                    [<option key="abcd" value="" disabled>Select Section</option>,
```
taimur38: instead of adding this to the beginning of the array, you can just add it outside the curl brackets first. then you don't have to make up a key

web/src/actions/index.js
```diff
@@ -406,4 +406,12 @@ export const addTag = (students, tag) => dispatch => {
 	}))
 
   	dispatch(createMerges(merges)) 
+}
+
+export const addDiary = (diary) => dispatch => {
+
+  	dispatch(createMerges([{
+		path: ["db", "diary"],
+		value : diary
```
taimur38: i have no idea what a diary is. how do i use this

taimur38: every time someone saves something to the diary it will mess up someone else doing it simultaneously

PR #110, opened 2019-03-10
Analytics improvements

raoaliahmad:


web/src/modules/Analytics/style.css
```diff
@@ -1,3 +1,25 @@
...
+	display:flex; 
+	flex-direction: row;
+	justify-content: center;
+}
+.chart-checkbox input{
+	height: 20px
+}
+.analytics .search-bar{
+	width: 100%;
+	height: 37px;
```
taimur38: this is a very strange height.

web/src/modules/Analytics/Attendance/index.js
```diff
@@ -114,29 +135,90 @@ export default connect(state => ({
...
+				/>
+				Leave
+			</div>
+			
+			<div className="chart-checkbox" style={{ color:"#ff6b68" }}>
+				<input
+					type="checkbox"
+					{...this.former.super_handle(["chartFilter", "percentage"])}
+				/>
+				Absentee (%)
```
taimur38: This should be wrapped in its own div, instead of just floating out here in between elements

web/src/modules/Analytics/Fees/index.js
```diff
@@ -147,28 +157,93 @@ class FeeAnalytics extends Component {
 		total_debts = { PAID: total_paid, OWED: total_owed, FORGIVEN: total_forgiven}
 	}
 
-	const items = Object.values(total_student_debts).filter(({student, debt}) => ( student.id && student.Name) && student.Name.toUpperCase().includes(this.state.filterText.toUpperCase()))
-				
+	const items = Object.values(total_student_debts)
+		.filter(({student, debt}) => (
+			student.id && student.Name) &&
```
taimur38: parenthesis should be on same line here

web/src/modules/Analytics/Fees/index.js
```diff
@@ -147,28 +157,93 @@ class FeeAnalytics extends Component {
...
+				/> 
+				Pending
+			</div>
+
+			<div className="chart-checkbox" style={{ color:"#74aced" }}>
+				<input
+					type="checkbox"
+					{...this.former.super_handle([ "chartFilter", "total" ])}
+				/> 
+				Total
```
taimur38: These should also all be wrapped in a div or something similar. fine for now

PR #111, opened 2019-03-11
Added Tags in student profile

raoaliahmad:


web/src/modules/Student/Single/Create/index.js
```diff
@@ -357,6 +365,50 @@ class SingleStudent extends Component {
...
+		const tags = new Set();
+
+		Object.values(this.props.students)
+			.forEach(s => {
+				Object.keys(s.tags || {})
+					.forEach(tag => tags.add(tag))
+			})
+
+		return tags;
+	}
```
taimur38: you don't need this method because the keys in an object are already unique

web/src/modules/Student/Single/Create/index.js
```diff
@@ -475,6 +527,30 @@ class SingleStudent extends Component {
...
+								</div>
+							)
+						}
+					</div>}
+
+					{!prospective && <div className="row" style={{flexDirection:"row"}}>
+						<input list="tags" onChange={(e) => this.setState({ new_tag: e.target.value })} placeholder="Type or Select Tag" style={{width: "initial"}}/>
+						<datalist id="tags">
+						{
+							[...this.uniqueTags().keys()]
```
taimur38: Object.keys(this.state.profile.tags) will do the same thing

web/src/modules/Student/Single/Create/index.js
raoaliahmad: this method is getting keys from all the students not just the current one and showing them as suggestions through data list, and there's a possibility that two different students might have the same tags.

PR #115, opened 2019-03-19
added filters in list component for students

raoaliahmad:


web/src/components/List/index.js
```diff
@@ -8,34 +9,129 @@ class List extends Component {
 	constructor(props) {
 		super(props)
 		this.state = {
-			filterText : ""
+			filterText : "",
+			active: true,
+			inActive: false,
```
taimur38: rename these to `showActiveStudent` and `showInactiveStudent`

PR #118, opened 2019-04-09
converted faculty and utils to typescript

raoaliahmad:


web/src/utils/localStorage.ts
```diff
@@ -171,7 +172,10 @@ const checkPermissions = state => {
 	state.db.settings = {
 		...state.db.settings,
 		permissions:{
-			fee:{ teacher: true }
+			...state.db.settings.permissions,
```
taimur38: this order seems incorrect, the state permissions will always be overridden by "true"

PR #119, opened 2019-04-12
Redesigned student list

raoaliahmad:
Sorry about the bad branch name.


web/src/components/Card/index.tsx
```diff
@@ -0,0 +1,74 @@
+import React,{Component} from 'react'
+import { Link, RouteComponentProps } from 'react-router-dom'
+import Former from '../../utils/former'
+
+import './style.css'
+
+interface P {
+	items: MISStudent[]
```
taimur38: This shouldn't be MISStudent either

PR #122, opened 2019-04-24
Certificates

raoaliahmad:


web/src/modules/Teacher/Single/Certificates/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+                <label>Type</label>
+                <select {...this.former.super_handle(["selectedCertificate"])}>
+                    <option value="EXPERIENCE">Experience</option>
+                </select>
+            </div>
+            <div className="button blue" onClick={()=> window.print()}>Print</div>
+          </div>
+          {
+            <this.getSelectedCertificate/>
+          }
```
taimur38: This is a weird combination of JSX and javascript. I think more conventional is if you either make a variable within the render function like `let Component = <ExperienceCertificate curr_teacher={this.teacher()} /> ......... <Component />` or if you just call `{ this.getSelectedCertificate() }` in render instead of putting the function inside jsx brackets

web/src/modules/Teacher/Single/Certificates/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+    <div className="head">
+      <div className="divider">EXPERIENCE CERTIFICATE</div>
+      <div className="sub-divider"> To Whom it May Concern</div>
+    </div>
+
+    <div className="body">
+      <div className="para">
+        This is to certify that <span style={{fontWeight:"bold", textDecoration:"underline"}}>{curr_teacher.Name}</span>, 
+        Son/Daughter of <span style={{fontWeight:"bold", textDecoration:"underline"}}>{curr_teacher.ManName ? curr_teacher.ManName : "________________"}</span>, has
```
taimur38: Do we not track the gender of the student in the app? We should be able to pick the correct word for Son or Daughter instead of writing "Son/Daughter"

web/src/modules/Teacher/Single/Certificates/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+        worked as a teacher of the following subjects in our school from <b style={{textDecoration:"underline"}}>{`${moment(curr_teacher.HireDate).format("DD-MM-YYYY")} `}</b>
+        to <b style={{textDecoration:"underline"}}>{moment(curr_date).format("DD-MM-YYYY")} </b>
+        </div>
+
+      <div className="cert-row">
+        <label>Subjects: </label>
+        <div></div>
```
taimur38: you could just write `<div />`

web/src/modules/Teacher/Single/Certificates/index.tsx
```diff
@@ -0,0 +1,135 @@
...
+      future endeavors.
+      </div>
+    </div>
+
+    <div className="footer">
+      <div className="left">
+        <div> Issuance Date</div>
+      </div>
+      <div className="right">
+        <div> Principal Signature</div>
```
taimur38: How come these don't need empty `<div />` elements next to them?

web/src/modules/Teacher/Single/Certificates/index.tsx
raoaliahmad: I think It's fine for now because if I use the correct gender I would need to check wherever we are using gender-specific terms, and that would simply mean a lot of checks. :D So, It's your call.

web/src/modules/Teacher/Single/Certificates/index.tsx
raoaliahmad: That's because they only have a top border, these are the right and Left signature Fields.

web/src/modules/Teacher/Single/Certificates/index.tsx
taimur38: write the code nicely

PR #125, opened 2019-05-08
Datesheet changes

raoaliahmad:


web/src/modules/Planner/style.css
```diff
@@ -13,7 +13,11 @@
 }
 
 .planner .section {
-	padding: 15px 0px;
+	padding: 15px 5px;
```
taimur38: section already has padding, why did this planner need it changed?

web/src/modules/Planner/style.css
```diff
@@ -63,11 +71,26 @@
...
 		padding: 15px 10px
 	}
+	.planner .row.input input{
+		width: 80%
+	}
+
+	.planner .row.input.info .row {
+		flex-direction: column;
+		align-items: baseline
+	}
```
taimur38: row inside a row?

web/src/modules/Planner/index.tsx
```diff
@@ -134,44 +195,77 @@ class Planner extends Component <propTypes, S> {
...
 						<option value="DATE_SHEET">Date Sheet</option>
 					</select>
-				</div>
-				<div className="divider">Date-Sheet</div>
+				</div> 
+			*/}
+				<div className="title">DateSheet</div>
 
-				<div> <b> Class/Section: </b> {`${curr_class.name +"/"+ curr_section.name}`}</div>
+				<div className="row input info">
```
taimur38: this doesn't seem right... we should review on monday

taimur38:
Should go over this on Monday

web/src/modules/Planner/style.css
raoaliahmad: because the section has a `15px` padding on its sides which seemed too much for date-sheet

web/src/modules/Planner/style.css
raoaliahmad: Yeah, it's a parent row and then there's is a child row inside it.

web/src/modules/Planner/index.tsx
raoaliahmad: okay

PR #127, opened 2019-05-29
Expense module

raoaliahmad:


web/src/modules/Expenses/General/index.tsx
```diff
@@ -0,0 +1,409 @@
...
+      .filter(t => t.Salary !== "")  
+      .forEach(t => {
+        const id = `${moment().format("MM-YYYY")}-${t.id}`
+        if(!this.props.expenses[id])
+        {
+          //id, amount, label, type, category, faculty_id
+          this.props.addSalaryExpense( id, parseFloat(t.Salary), this.props.teachers[t.id].Name, "PAYMENT_DUE", "SALARY", t.id )
+          console.log("<================================>\n")
+        } 
+      }) */
```
taimur38: When doing automatic, instead of saving each salary expense individually see if it is possible to make it into a single action that saves all the new expenses

web/src/modules/Expenses/General/index.tsx
```diff
@@ -0,0 +1,409 @@
...
+          active: true,
+          good: false,
+          text: erroText
+        }
+      })
+    }
+
+    if(payment.category === "SALARY"){
+
+      this.props.addSalaryExpense( id, parseFloat(payment.amount), this.props.teachers[payment.faculty_id].Name, "PAYMENT_GIVEN", payment.category, payment.faculty_id, payment.date,0,parseFloat(payment.deduction))
```
taimur38: should make sure that the parseFloats don't return NaN (not a number)

web/src/modules/Expenses/General/index.tsx
```diff
@@ -0,0 +1,409 @@
...
+            active: false
+          }
+        })
+      }, 1000)
+
+      this.newPayment()
+      return
+    }
+
+    this.props.addExpense( parseFloat(payment.amount), payment.label, "PAYMENT_GIVEN", payment.category, parseFloat(payment.quantity), payment.date)
```
taimur38: also here, should make sure the parseFloats are returning proper numbers

web/src/modules/Expenses/General/index.tsx
```diff
@@ -0,0 +1,409 @@
...
+
+    }
+    //when both are not empty
+    if(month !== "" && year !== "")
+    {
+      return moment(payment.date).format("MMMM") === month && moment(payment.date).format("YYYY") === year;
+    }
+  }
+
+  addSalary = (e: { target: { value: any; }; }) => {
```
taimur38: shouldn't have semicolons here, not sure how it's not a syntax error

taimur38: also this function is too similar to addSalaryExpense, a more descriptive name, esp since you're not using former, would be onTeacherSelect

web/src/modules/Expenses/General/index.tsx
```diff
@@ -0,0 +1,409 @@
...
+    
+    const total_filtered_expense = Object.values(expenses)
+      .filter(e => this.getFilterCondition(this.state.yearFilter,this.state.monthFilter,e))
+      .reduce((agg, curr) => {
+        if(curr.type === "PAYMENT_GIVEN"){
+          return agg + curr.amount - (curr.expense === "SALARY_EXPENSE" ? curr.deduction : 0)
+        }
+        else
+          return agg 
+      },0)
```
taimur38: this and the previous total expense can be put into a single reduce. you will just have to have an if statement inside the reduce `if(this.getFilterCondition(e)) { etc }`

If it doesn't make the code too hard to read, you should be thinking about what happens when there is 5 years of expense data and they're using the worst phone of all time. The app will get slow

taimur38: Although even then, I doubt it would get slow. but still.

web/src/modules/Expenses/General/index.tsx
```diff
@@ -0,0 +1,409 @@
...
+
+      this.newPayment()
+      return
+    }
+
+    this.props.addExpense( parseFloat(payment.amount), payment.label, "PAYMENT_GIVEN", payment.category, parseFloat(payment.quantity), payment.date)
+    this.newPayment()
+  }
+
+  getFilterCondition = (year: string, month: string, payment: any) =>
```
taimur38: if year and month are getting pulled from state, do we need them in the signature?

web/src/modules/Expenses/General/index.tsx
```diff
@@ -0,0 +1,409 @@
...
+              <option value="RENT">Rent</option>
+              <option value="ACTIVITY">Student Activity</option>
+              <option value="DAILY">Daily</option>
+              <option value="PETTY_CASH">Petty Cash</option>
+            </select>
+          </div>
+
+          { this.state.payment.category === "SALARY" && <div className="row">
+              <label> Teacher </label>
+              <select {...this.former.super_handle(["payment", "faculty_id"])} onChange={(e) => this.addSalary(e)}>
```
taimur38: you should be careful overriding formers onChange method. do you need former at all for this entry?

web/src/modules/Expenses/Income-Expenditure/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+		},
+		monthFilter: "",
+		yearFilter: ""
+
+    }
+
+    this.former = new Former (this,[])
+  }
+
+  getFilterCondition = (year: string, month: string, payment: any) =>
```
taimur38: if this function is being copy pasted everywhere it should go into utils

web/src/modules/Expenses/Income-Expenditure/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+  .filter(([id,e]) => e.type === "PAYMENT_GIVEN")
+  .reduce((agg, [id, curr]) => {
+    
+    return {
+      ...agg,
+      [id]: curr
+    }
+  }, {})
+
+  const income_exp = {...stu_payments, ...filtered_expense} as MISStudentPayment | MISExpense | MISSalaryExpense
```
taimur38: same comment as this block of code on the other module. merge the reduces, and put a type on the default value so you dont need to type-cast this merged object.

web/src/modules/Expenses/Income-Expenditure/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+    Months.add(moment(s.date).format("MMMM"))
+    Years.add(moment(s.date).format("YYYY"))
+  }
+
+  const income_exp_sorted = Object.values(income_exp)
+    .filter(e => this.getFilterCondition(this.state.yearFilter, this.state.monthFilter, e))
+    .sort((a, b) => a.date - b.date)
+
+  let total_income = 0
+  let total_expense = 0
```
taimur38: if all this code is the same between the expense module components, should there be a higher-level parent component that does all this and then these can be smaller components that just use from props?

web/src/modules/Expenses/Income-Expenditure/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+            
+            <div className="table row">
+              <label>Total Expense:</label>
+               <div><b>{numberWithCommas(total_monthly_expense)}</b></div>
+            </div>
+
+            <div className="divider">Ledger</div>
+
+            <div className="filter row no-print" style={{marginBottom:"10px"}}>
+              <select className="" {...this.former.super_handle(["monthFilter"])} style={{ width: "150px" }}>
```
taimur38: why is there a specific pixel width on this one

web/src/modules/Expenses/Income-Expenditure/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+              {
+                Object.entries(income_exp_sorted)
+                .map(([id,e]) => {
+                    return <div key={id} className="table row">
+                      <label> {moment(e.date).format("DD-MM-YYYY")} </label>
+                      <label> {e.label? e.label : e.type === "SUBMITTED" ? "PAID": "-" }</label>
+                      <label> {e.category ? e.category : e.fee_name || "-"}</label>
+                      <label> {e.quantity ? e.quantity : "1"} </label>
+                      <label> {e.type === "PAYMENT_GIVEN" ? -1 * (e.amount - ((e.expense === "SALARY_EXPENSE" ? e.deduction : 0))) : e.amount}</label>
```
taimur38: extra parenthesis

web/src/modules/Expenses/Income-Expenditure/index.tsx
```diff
@@ -0,0 +1,232 @@
...
+                <label><b> Amount </b></label>
+              </div>
+
+              {
+                Object.entries(income_exp_sorted)
+                .map(([id,e]) => {
+                    return <div key={id} className="table row">
+                      <label> {moment(e.date).format("DD-MM-YYYY")} </label>
+                      <label> {e.label? e.label : e.type === "SUBMITTED" ? "PAID": "-" }</label>
+                      <label> {e.category ? e.category : e.fee_name || "-"}</label>
```
taimur38: can also do `e.category || e.fee_name || "-"`

web/src/modules/Expenses/style.css
```diff
@@ -0,0 +1,21 @@
...
+    width: 100%;
+    overflow-x: hidden;
+}
+
+.expenses {
+	width: 90%;
+}
+
+.expenses .form {
+	display: -webkit-flex;
```
taimur38: you dont need to put these prefixes in. When the app is compiled it is automatically added

web/src/modules/Analytics/Expenses/index.tsx
```diff
@@ -0,0 +1,291 @@
...
+		collective_obj: { [month: string]: { income: number, expense: number }}
+		total_income: number 
+		total_expense: number
+	}
+	
+	const MonthlyExpenseTable: React.SFC<TableProps> = ({ collective_obj, total_income, total_expense }) => {
+	
+		return <div className="section table" style={{margin: "20px 0", backgroundColor:"#c2bbbb21", overflowX: "scroll" }}>
+				<div className="table row heading">
+					<label style={{ backgroundColor: "#efecec", textAlign:"center" }}> <b> Date     </b></label>
```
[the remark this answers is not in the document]
raoaliahmad: Nope, I think that was added accidentally.

web/src/actions/index.js
```diff
@@ -347,6 +347,69 @@ export const addMultiplePayments = (payments) => dispatch => {
...
+				type,
+				category,
+				quantity,
+				date
+			}
+		}
+	]))
+}
+
+export const addSalaryExpense = (id, amount, label, type, category = "SALARY", faculty_id, date, advance, deduction ) => dispatch => {
```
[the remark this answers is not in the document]
raoaliahmad: Oh my bad.

web/src/modules/Expenses/Income-Expenditure/index.tsx
raoaliahmad: It's only being used in two places. If I need it more I'll move it.

web/src/modules/Expenses/General/index.tsx
raoaliahmad: I didn't use it from the state because I was thinking of adding it to utils but I didn't.

web/src/modules/Expenses/General/index.tsx
```diff
@@ -4,406 +4,517 @@ import Former from '../../../utils/former';
...
+						<input type="number" {...this.former.super_handle(["payment", "amount"])} placeholder="Enter Amount" />
+					</div>
+					{this.state.payment.category !== "SALARY" && <div className="row">
+						<label>Quantity</label>
+						<input type="number" {...this.former.super_handle(["payment", "quantity"])} placeholder="Enter Quantity" />
+					</div>}
+					{this.state.payment.category === "SALARY" && <div className="row">
+						<label>Deductions</label>
+						<input type="number" {...this.former.super_handle(["payment", "deduction"])} placeholder="If any" />
+					</div>}
```
taimur38: Instead of putting all these conditionals on `category !== 'SALARY'` you should consider making this its own component. You would pass former in as a prop. but this is fine for now

PR #136, opened 2019-07-09
Added Historical Fee, updated annual charges and updated timeout to 30s

raoaliahmad:


web/src/modules/Settings/historical-fee.tsx
```diff
@@ -0,0 +1,231 @@
...
+		this.former = new former(this, [])
+	}
+
+	save = () => {
+		
+		const payment = {
+			...this.state.fee,
+			amount_owed: parseFloat(this.state.fee.amount_owed) || 0,
+			amount_paid: parseFloat(this.state.fee.amount_paid) || 0,
+			amount_forgiven: parseFloat(this.state.fee.amount_forgiven) || 0
```
taimur38: ah I see we are parseFloating here

web/src/modules/Settings/historical-fee.tsx
```diff
@@ -0,0 +1,231 @@
...
+			
+		const selected_student = students[this.state.selected_student]
+		
+		const filteredPayments = selected_student ? getFilteredPayments(selected_student, this.state.yearFilter, this.state.monthFilter) : false
+		const curr_class_name = this.state.selected_class ? class_Items.find( s => s.id === this.state.selected_class).namespaced_name : "None Selected"
+		
+		const Months = new Set()
+		const Years = new Set()
+
+		if(selected_student){
```
taimur38: What happens if they selected a class?

web/src/modules/Settings/style.css
```diff
@@ -67,3 +67,13 @@ div.settings .new-badge {
...
+
+/*<==============> Historical Fee Css <=============>*/
+
+.historical-fees .section {
+	width: 80%;
+}
+
+.historical-fees.form {
+	padding-bottom: 0px !important;
+}
```
taimur38: the structure under settings should be another folder called historical fees with its own index.tsx and a style.css with it.

web/src/modules/Settings/historical-fee.tsx
raoaliahmad: if the selected student is in that class then nothing, otherwise they would have to reselect the student.

PR #139, opened 2019-08-02
Certificates two

Mudassar045:


web/src/components/Layout/index.js
```diff
@@ -25,13 +25,17 @@ const Header = ({user, history}) => <div className="header">
...
-				<div className="logo-container" style={{width: "20%"}}>
-					{logo !== "" && <img className="header-logo" src={logo} alt="No Logo"/>}
+				<div className="logo-container" style={{width: "20%", objectFit:"center"}}>
+					{logo !== "" && <img className="header-logo" src={logo} alt="School Logo"/>}
 				</div>
 				<div className="header-style">
-					<div className="title">{settings.schoolName}</div>
-					<div className="address" style={{marginBottom: "5px"}}>{settings.schoolAddress}</div>
-					<div className="phone-number">{settings.schoolPhoneNumber}</div>
+					<div className="title">{settings.schoolName.toUpperCase() || ""}</div>
```
taimur38: this doesn't handle the case if schoolName is undefined.

web/src/components/Layout/index.js
```diff
@@ -25,13 +25,17 @@ const Header = ({user, history}) => <div className="header">
 
 export const PrintHeader = ({settings, logo}) => <div className="print-only school-header">
 			<div className="header-body">
-				<div className="logo-container" style={{width: "20%"}}>
-					{logo !== "" && <img className="header-logo" src={logo} alt="No Logo"/>}
+				<div className="logo-container" style={{width: "20%", objectFit:"center"}}>
```
taimur38: don't think objectFit is needed here

web/src/modules/Student/List/index.js
```diff
@@ -29,6 +29,11 @@ const StudentItem = (S) => {
 						<div className="para-row">
 							<b>{`Adm #: `}</b>{S.AdmissionNumber}
 						</div>}
+					{
+						<div className="para-row">
+							<b>{`Phone:`}</b>{S.Phone?S.Phone.substr(0,10):""}
```
taimur38: put spaces in between these commands. 
` S.Phone ? S.Phone.substr(0, 10) : ""`

also why do we need this substring.

web/src/modules/Student/Single/Certificates/style.css
```diff
@@ -101,9 +101,20 @@
     @media print {
         .certificate {
             margin-top: 50px;
-            border: 10px double rgba(0,0,0,0.3);
```
taimur38: it should have a border. almost every example certificate had a border around it

web/src/components/Layout/index.js
Mudassar045: Acknowledged, I'll apply validations here

web/src/components/Layout/index.js
Mudassar045: ok, I will remove it

PR #140, opened 2019-08-07
Actions typescript

Mudassar045:


web/src/actions/index.ts
```diff
@@ -0,0 +1,737 @@
...
+ 	dispatch(createDeletes([
+		{
+			path: ["db", "exams", exam_id]
+		},
+		...deletes
+	]))
+
+}
+
+type History = {
```
taimur38: use `MISSMSHistory` from typings.d.ts

but note that it doesn't look like this has all the correct fields. For example, logSms in the Help component also adds faculty fields, and text as well. 

we should also have a fixed amount of "TYPE" variables not just a string. this will take some searching around the codebase to consolidate

web/src/actions/index.ts
```diff
@@ -0,0 +1,737 @@
...
+	//logo_string is a base64 string
+	dispatch(createMerges([
+		{
+			path : ["db", "assets", "schoolLogo"],
+			value: logo_string
+		}
+	]))
+}
+
+export const addDiary = (section_diary: { }, section_id: string) => (dispatch: Function) => {
```
taimur38: here you are saying the section_diary can only be an empty object, which is definitely incorrect. 

and it turns out diary is not even accounted for in typings.d.ts so it will need to be added there

web/src/actions/index.ts
```diff
@@ -0,0 +1,737 @@
...
+		  path:["db", "diary", "date"],
+		  value: new Date().getTime()
+	  }
+	])) 
+}
+
+type Payment = {
+	[id: string]:{
+		amount: number
+		fee_id?: string
```
taimur38: fee_id can not be undefined in this

taimur38: and the type should not be called Payment - it should be more specific. something like `EditPaymentItem` would be more appropriate

taimur38: also the signature of this function in the Fees object says that payments is of type `MISStudent["payments"]` so this is also incorrect

web/src/actions/index.ts
```diff
@@ -0,0 +1,737 @@
...
+		"D": string
+		"F": string
+	}	
+}
+
+type Devices = {
+	[id: string] : string
+}
+
+type Settings = {
```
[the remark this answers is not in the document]
Mudassar045: Got it.

web/src/actions/index.ts
```diff
@@ -0,0 +1,737 @@
...
+	namespaced_name: string
+	className: string
+	classYear: number
+	name: string
+	faculty_id?: string
+}
+
+export const promoteStudents = (promotion_map: PromotionMap, section_metadata: Section[]) => (dispatch: Function) => {
+
+	// accept a map of key: student_id, value: new section_id
```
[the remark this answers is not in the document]
Mudassar045: I didn't add this, but I'll update it

web/src/actions/index.ts
```diff
@@ -0,0 +1,737 @@
...
+
+	// think about the case when someone promotes up and down repeatedly. 
+	// this will overwrite their history... instead of adding to it.
+
+
+	const merges = Object.entries(promotion_map).reduce((agg, [student_id, {current, next}]) => {
```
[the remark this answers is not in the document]
Mudassar045: exactly, I looked here and updated the type

web/src/actions/index.ts
```diff
@@ -0,0 +1,737 @@
...
+export const SIGN_UP_LOADING = "SIGN_UP_LOADING"
+export const SIGN_UP_SUCCEED = "SIGN_UP_SUCCEED"
+export const SIGN_UP_FAILED = "SIGN_UP_FAILED"
+
+type Profile = {
+	name: string
+	phone: string
+	city: string
+	schoolName: string
+	packageName: string
```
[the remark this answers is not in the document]
Mudassar045: in Signup.js
packageName: "Free-Trial" is defined. Now what I do? make there too or just mention "Free-Trial" in actions file?

web/src/actions/index.ts
```diff
@@ -0,0 +1,737 @@
...
+export const undoFacultyAttendance = (faculty: MISTeacher, date: string) => (dispatch: Function) => {
+	dispatch(createDeletes([
+		{
+			path:["db", "faculty", faculty.id, "attendance", date]
+		}
+	]))
+} 
+
+
+export const addPayment = (student: MISStudent, payment_id: string, amount: number, date: number, type: string, fee_id?: string, fee_name?: string) => (dispatch: Function) => {
```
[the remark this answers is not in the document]
Mudassar045: Yes, I did.  I will revert it

web/src/actions/index.ts
```diff
@@ -0,0 +1,696 @@
...
+			}
+		})
+	}
+
+	dispatch(createMerges(merges))
+}
+
+interface ExpenseAddItemCategory {
+	category: "SALARY" | "BILLS" | "STATIONERY" | "REPAIRS" | "RENT" | "ACTIVITY" | "DAILY" | "PETTY_CASH"   
+}
```
taimur38: don't need this anymore

web/src/actions/index.ts
```diff
@@ -0,0 +1,696 @@
...
+
+interface FeeAddItem{
+	student:MISStudent 
+	fee_id: string
+	amount: number
+	type: "FEE" | "SCHOLARSHIP" | ""
+	period: "MONTHLY" | "SINGLE" | ""
+	name: string
+}
```
taimur38: Should pull anything you can from typings.d.ts using MISStudentFee
```typescript
type FeeAddItem = MISStudentFee & {
    student: MISStudent
    fee_id: string
}
```

web/src/actions/index.ts
taimur38: also double check the count is a number or a string

web/src/actions/index.ts
```diff
@@ -0,0 +1,737 @@
...
+	//Id of the expense to be deleted
+
+	dispatch(createDeletes([
+		{
+			path: ["db", "expenses", id]
+		}
+	]))
+}
+
+export const addMultipleFees = (fees: { student:MISStudent, fee_id: string, amount: number, type: string, period: string, name: string}[]) => (dispatch: Function) => {
```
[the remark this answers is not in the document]
Mudassar045: Yes I double-checked and converted like this
```
type FeeAddItem  = MISStudentFee & {
	student: MISStudent 
	fee_id: string
}
```

web/src/actions/index.ts
Mudassar045: I double-checked, Yes **count** is a number

PR #151, opened 2019-09-02
Family ID

taimur38:
**Features:**
- [ ] FamilyID code added to each student
- [ ] Families can be inspected from family module
- [ ] Ledgers are combined
- [ ] Analytics treats families as single unit 

**Missing**
- [ ] Ability to add new families from family module (create new family)
- [ ] Manage fees for whole family from family module

### DIFF
--- a/web/src/modules/Student/Single/Fees/index.tsx
+++ b/web/src/modules/Student/Single/Fees/index.tsx
@@ -117,6 +117,40 @@ class StudentFees extends Component <propTypes, S> {
 		return this.props.students[id];
 	}
 
+	siblings = () => {
+
+		const student = this.student();
+
+		return Object.values(this.props.students)
+			.filter(s => s.Name && s.FamilyID && s.FamilyID === student.FamilyID)
+
+	}
+
+	mergedPayments = () => {
+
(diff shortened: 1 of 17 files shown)


web/src/modules/Student/Single/Fees/index.tsx
```diff
@@ -117,6 +117,40 @@ class StudentFees extends Component <propTypes, S> {
 		return this.props.students[id];
 	}
 
+	siblings = () => {
```
raoaliahmad: should move this to utils

web/src/modules/Student/Single/Fees/index.tsx
```diff
@@ -117,6 +117,40 @@ class StudentFees extends Component <propTypes, S> {
...
+
+		const student = this.student();
+
+		return Object.values(this.props.students)
+			.filter(s => s.Name && s.FamilyID && s.FamilyID === student.FamilyID)
+
+	}
+
+	mergedPayments = () => {
```
raoaliahmad: Are you sure it's a good idea to show siblings payments while viewing a students. Shouldn't there be a separate page to view siblings fee record

PR #160, opened 2019-09-20
Indexed db

raoaliahmad:
Dont Merge it, Still needs more work. Just read the code.

Check list;
- [ ] Read the code


web/src/modules/Login/index.js
```diff
@@ -40,9 +41,23 @@ class Login extends Component {
...
+		}).then(db => {
+			db.delete('root-state', 'db')
+				.then(res => {
+					this.props.history.push("/landing")
+					window.location.reload()
+				})
+				.catch(err => console.error(err))
+		})
+		.catch(err => console.error(err))
+
```
taimur38: there should be a util function to get access to the DB instead of using this code everywhere

web/src/modules/Settings/index.js
```diff
@@ -332,17 +333,44 @@ class Settings extends Component {
...
+				upgrade(db) {
+					db.createObjectStore('root-state')
+				}
+			})
+			
+			if (db) {
+				console.log("Exporting From idb")
+				const state = await db.get('root-state', 'db')
+				const a = document.createElement("a")
+				const client_id = localStorage.getItem("client_id")
```
taimur38: here you should just connect the settings component to the db and load it from memory instead of again loading it from the backend

PR #161, opened 2019-09-24
Clickable stats attach to attendance analytics

Mudassar045:


web/src/modules/Analytics/Attendance/index.tsx
```diff
@@ -123,48 +124,67 @@ interface S {
 	},
 	classFilter: string,
 	selected_section_id: string,
-	selected_period: string,
+	selected_period: string | String [],
```
taimur38: why would we have an array of selected periods?

web/src/modules/Analytics/Attendance/index.tsx
```diff
@@ -123,48 +124,67 @@ interface S {
...
-		 start_date: moment().subtract(1,'year').unix() * 1000,
-		 end_date: moment().unix() * 1000,
-		 isStudentAttendanceFilter: false,
-	  }
-	  this.former = new Former(this, [])
+		super(props)
+		 
+		const parsed_query = queryString.parse(this.props.location.search);
+
+		const start_date = parsed_query["?start_date"] || ""
```
taimur38: there shouldn't be question marks here

web/src/modules/Analytics/Attendance/index.tsx
```diff
@@ -123,48 +124,67 @@ interface S {
...
+				present: true,
+				absent: true,
+				leave: true,
+				percentage: true
+			},
+			classFilter: "",
+			selected_section_id: "",
+			selected_period: period !== "" ? period : "Monthly",
+			start_date: start_date !== "" ? moment(start_date, "MM-DD-YYYY").unix() * 1000 : moment().subtract(1,'year').unix() * 1000,
+			end_date: end_date !== "" ? moment(end_date, "MM-DD-YYYY").unix() * 1000 : moment().unix() * 1000,
```
taimur38: this logic should take place where you are initializing the variable up top instead of here in your state definition

web/src/modules/Analytics/Attendance/index.tsx
```diff
@@ -123,48 +124,67 @@ interface S {
...
+			search: `?start_date=${moment(this.state.start_date).format("MM-DD-YYYY")}&end_date=${moment(this.state.end_date).format("MM-DD-YYYY")}&period=${this.state.selected_period}`
+		})
+
+	}
+
+	componentWillReceiveProps(nextProps : propTypes) { 
+
+		const parsed_query = queryString.parse(nextProps.location.search);
+
+		const start_date = parsed_query["?start_date"] || ""
```
taimur38: shouldn't need question marks here

web/src/modules/Analytics/Attendance/index.tsx
```diff
@@ -349,12 +369,14 @@ class AttendanceAnalytics extends Component < propTypes, S > {
...
 			<div className="table row">
 				<label><b>Name</b></label>
+				<label><b>Phone</b></label>
 				<label><b>Days Absent</b></label>
 			</div>
 			{
 				items
 					.map(([ sid, { student, PRESENT, ABSENT, LEAVE } ]) => <div className="table row">
 						<Link to={`/student/${sid}/attendance`}>{student.Name}</Link>
+						<div>{student.Phone}</div>
```
taimur38: wasn't this supposed to be on print only?

web/src/modules/Analytics/Teacher-Attendance/index.tsx
```diff
@@ -75,23 +75,23 @@ const MonthlyAttendanceTable = ({monthly_attendance, totals}: TableData) =>{
...
+						.sort(([month, ], [m2, ]) => moment(month).isBefore(m2) ? 1 : 0)
 						.map(([month, {PRESENT, LEAVE, ABSENT} ]) =>
 						
 							<div className="table row" key={Math.random()}>
 								<div style={{ backgroundColor: "#efecec"}}>{month }</div>
 								<div style={{ backgroundColor: "#93d0c5"}}>{PRESENT}</div>
 								<div style={{ backgroundColor: "#fc6171"}}>{ABSENT}</div>
 								<div style={{ backgroundColor: "#e0e0e0"}}>{LEAVE}</div>
-								<div style={{ backgroundColor: "#bedcff"}}>{ isFinite( Math.round((ABSENT / (PRESENT + LEAVE )) * 100)) ? Math.round((ABSENT / (PRESENT + LEAVE )) * 100) : "0"}%</div>
+								<div style={{ backgroundColor: "#bedcff"}}>{ isFinite(ABSENT / (ABSENT + PRESENT + LEAVE) * 100) ? (ABSENT / (ABSENT + PRESENT + LEAVE) * 100).toFixed(2) : "0"}%</div>
```
taimur38: why do we have to check isFinite?

web/src/modules/Analytics/Teacher-Attendance/index.tsx
```diff
@@ -75,23 +75,23 @@ const MonthlyAttendanceTable = ({monthly_attendance, totals}: TableData) =>{
...
 							</div>
 						),
 						<div className="table row footing" style={{borderTop: '1.5px solid #333'}} key={Math.random()}>   
 							<label style={{ backgroundColor: "#efecec"}}><b>Total</b></label>
 							<label style={{ backgroundColor: "#93d0c5"}}><b>{totals.PRESENT}</b></label>
 							<label style={{ backgroundColor: "#fc6171"}}><b>{totals.ABSENT}</b></label>
 							<label style={{ backgroundColor: "#e0e0e0"}}><b>{totals.LEAVE}</b></label>
-							<label style={{ backgroundColor: "#bedcff"}}><b>{Math.round((1 - totals.ABSENT / (totals.PRESENT + totals.LEAVE)) * 100)}%</b></label>
+							<label style={{ backgroundColor: "#bedcff"}}><b>{ (totals.ABSENT / (totals.ABSENT + totals.PRESENT + totals.LEAVE) * 100).toFixed(2)}%</b></label>
```
taimur38: again we are removing the 1 - %. are we consistent about this, and why are we changing?

web/src/modules/Analytics/Attendance/index.tsx
```diff
@@ -91,15 +92,15 @@ const AttendanceTable = ({attendance, totals}: TableData) =>{
...
+								<div style={{ backgroundColor: "#bedcff"}}>{(ABSENT / ( ABSENT + PRESENT + LEAVE + CASUAL_LEAVE + SHORT_LEAVE + SICK_LEAVE) * 100).toFixed(2)}%</div>
 							</div>
 						),
 						<div className="table row footing" style={{borderTop: '1.5px solid #333'}} key={Math.random()}>   
 							<label style={{ backgroundColor: "#efecec"}}><b>Total</b></label>
 							<label style={{ backgroundColor: "#93d0c5"}}><b>{totals.PRESENT}</b></label>
 							<label style={{ backgroundColor: "#fc6171"}}><b>{totals.ABSENT}</b></label>
 							<label style={{ backgroundColor: "#e0e0e0"}}><b>{totals.LEAVE + totals.CASUAL_LEAVE + totals.SHORT_LEAVE + totals.SICK_LEAVE}</b></label>
-							<label style={{ backgroundColor: "#bedcff"}}><b>{Math.round((1 - totals.ABSENT / (totals.PRESENT + (totals.LEAVE + totals.CASUAL_LEAVE + totals.SHORT_LEAVE + totals.SICK_LEAVE))) * 100)}%</b></label>
+							<label style={{ backgroundColor: "#bedcff"}}><b>{(totals.ABSENT / (totals.ABSENT + totals.PRESENT + totals.LEAVE + totals.CASUAL_LEAVE + totals.SHORT_LEAVE + totals.SICK_LEAVE) * 100).toFixed(2)}%</b></label>
```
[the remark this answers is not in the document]
Mudassar045: both ways are approximately correct for calculating absentee %. but the just % way is pretty straightforward. 
For the first way, we are saying that 100% absentee and just subtract some % value of *present + leave* students. In the second way, we're calculating based on actual value of the expression.

web/src/modules/Analytics/Attendance/index.tsx
Mudassar045: we're using an old version of queryString, which is not correctly parsing the query URL so the parsing operation doesn't separate "?" from the first param. In new queryString v5, this thing automatically handle.

web/src/modules/Analytics/Attendance/index.tsx
Mudassar045: I was confused too here. QueryString parse method returns object of params, which can be accessible like parse_query_url.period (will give string ), but due to the first param, (I can't do this -> parse_query_url.?start_date), to avoid consistency, I used parse_query_url["period"], which returning String [], for the case of start and end date, moment(), handles automatically

web/src/modules/Analytics/Attendance/index.tsx
Mudassar045: Acknowledged

web/src/modules/Analytics/Attendance/index.tsx
Mudassar045: the reason already stated in the above reply

web/src/modules/Analytics/Attendance/index.tsx
Mudassar045: Like for fee_defaulter (in fee analytics), principal or the school owner can make call directly to the student having more absences

web/src/modules/Analytics/Teacher-Attendance/index.tsx
Mudassar045: Little ignorance, the result of (ABSENT / ABSENT + PRESENT + LEAVE ) will always be a number. I will remove it.

web/src/modules/Analytics/Teacher-Attendance/index.tsx
Mudassar045: the same reason which is stated above.

Mudassar045: Yes, the result of the expression is consistent.

web/src/modules/Analytics/Attendance/index.tsx
Mudassar045: https://github.com/sindresorhus/query-string

web/src/modules/Analytics/Attendance/index.tsx
taimur38: ok

web/src/modules/Analytics/Teacher-Attendance/index.tsx
taimur38: yes but it should still be rounded

taimur38: ok will resolve when code is changed

web/src/modules/Analytics/Attendance/index.tsx
taimur38: query string will give an array because the url could be `?value=1&value=2&value=3` so in this case value should be an array.

but in your case you know that you will only consider one. so make it into a single item and dont set state to the array and change this typing back to just string because it makes no sense to look at this state type and have selected_period be a string[]

web/src/modules/Analytics/Attendance/index.tsx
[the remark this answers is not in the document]
taimur38: you fixed the problem with the denominator. but then you went and changed what number we are displaying entirely. You are showing now the complete opposite number of what we have always been showing in the app. There should be a good reason for why you are switching. But you are not giving me any justification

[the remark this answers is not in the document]
Mudassar045: Here's justification in form of simple math
Consider the following scenario
There are 10 attendance actions (A=2, P=6, L=2)
Now result with the previous formula (no. what we have always in the app)
 Absentee % = 1 - 2/(6+2)*100 = 75% (incorrect)
Now I add A = 2 (in the denominator)
Absentee % = 1 - 2/(2+6+2)*100 = 80% (incorrect)
Now, in my expression (which gives complete opposite no. )
Absentee % = 2 / (2 + 6 + 2) *100 = 20% (correct)
Now I can hope, this simple math would be a good reason for switching

web/src/modules/Analytics/Attendance/index.tsx
Mudassar045: Acknowledged and will do that

PR #162, opened 2019-09-24
save multiple diary

Mudassar045:
Please read the code and logic
- [ ] read
Note: there is  an issue, to update single subject diary against section id, will look for the solution

### DIFF
--- a/web/src/modules/Diary/index.js
+++ b/web/src/modules/Diary/index.js
@@ -168,12 +168,12 @@ class Diary extends Component {
 
 	getSelectedSectionName = () => getSectionsFromClasses(this.props.classes)
 										.filter(s => s.id === this.state.selected_section_id)
-										.map (s => s.namespaced_name)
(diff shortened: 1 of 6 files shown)


web/src/modules/Diary/index.js
```diff
@@ -11,56 +11,23 @@ import former from 'utils/former'
...
- *    ...
- * }
- */
 	  
 class Diary extends Component {
 	constructor(props) {
 		super(props)
 
-		const curr_date = moment().format("DD/MM/YYYY")
-		const diary = Object.values(this.props.classes)
```
raoaliahmad: you should add a template but only for current class, and if he changes nay other sections,or dates diary we can just append the changes to state on call back

Mudassar045: I don't why there is a need to pre-store diary structure in database for current selected section. Can you explain the logic behind it?

PR #166, opened 2019-09-27
General sort apply to selectors/lists, teacher certificates editable and minor fixes

Mudassar045:
This PR contains followings updates:
- [x] Sorting Teachers name
- [x] Sorting Classes name
- [x] Sorting Students name
- [x] Ledger fixes
- [x] Teachers editable certificates


web/src/modules/Class/Single/Create/index.js
```diff
@@ -305,6 +306,7 @@ class SingleClass extends Component {
 										<option disabled selected value>select teacher</option>
 										{
 											Object.values(this.props.faculty)
+												.sort((a, b) => a.Name.localeCompare(b.Name))
```
taimur38: should filter for name first

web/src/modules/Teacher-Attendance/index.js
```diff
@@ -52,6 +52,7 @@ class TeacherAttendance extends Component {
 				{
 					Object.values(this.props.faculty)
 						.filter(f => f.Active)
+						.sort((a, b) => a.Name.localeCompare(b.Name))
```
taimur38: filter for name being defined

web/src/modules/Teacher-Attendance/index.js
```diff
@@ -51,7 +51,7 @@ class TeacherAttendance extends Component {
 				<div className="list">
 				{
 					Object.values(this.props.faculty)
-						.filter(f => f.Active)
+						.filter(f => f && f.Active && f.Name !=="")
```
taimur38: just check f.Name instead of `f.Name !== "" `

just checking if it is an empty string doesn't solve our undefined problem. Name could still be undefined and this would throw an error. you can test in javascript that both undefined and empty string resolve to `false` - so just do 

`f && f.Active && f.Name`

Mudassar045: Acknowledged!
I checked the create new teacher form, teacher name is mandatory, so `f.Name` cannot be `undefined` or `empty string`

taimur38: we check name because there can be combinations of events that happen over syncing which leave students not deleted properly. They will always be missing the "Name" field, so we check it

PR #170, opened 2019-10-15
add filters in analytics

Mudassar045:
Here's what this PR contains
- [x] add filters in teacher-attendance
- [x] add filters in expense
- [x] attach `daily stats` with analytics (teacher-attendance)
- [x] fix issues
- [x] code refactor

### DIFF
--- a/web/src/modules/Analytics/Expenses/index.tsx
+++ b/web/src/modules/Analytics/Expenses/index.tsx
@@ -21,7 +21,7 @@ interface ChartProps {
 	date_format: string
 }
 
-	const MonthlyExpenseChart : React.SFC <ChartProps> = ({collective_obj, chartFilter, date_format}) => {
+	const ExpenseChart : React.SFC <ChartProps> = ({collective_obj, chartFilter, date_format}) => {
 		return <ResponsiveContainer width="100%" height={200}>
 			<LineChart
 				data={
@@ -168,7 +168,7 @@ class ExpenseAnalytics extends Component<propTypes, S> {
 
 	const {students, expenses ,settings, schoolLogo} = this.props
 
-	const stu_payments = Object.entries(students)
+	const students_payments = Object.entries(students)
 		.filter(([id, s]) => s.Name)
 		.reduce((prev,[id, s]) => {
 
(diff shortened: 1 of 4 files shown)


web/src/modules/Analytics/Expenses/index.tsx
```diff
@@ -111,25 +116,64 @@ class ExpenseAnalytics extends Component<propTypes, S> {
...
 	}
 
 	calculateDebt = ({SUBMITTED, FORGIVEN, OWED, SCHOLARSHIP}:{ SUBMITTED: number, FORGIVEN: number, OWED : number, SCHOLARSHIP : number}) => SUBMITTED + FORGIVEN + SCHOLARSHIP - OWED;
 
+	onStateChange = () => {
+
+		this.props.history.push({
+			pathname: '/analytics/expenses',
+			search: `?start_date=${moment(this.state.start_date).format("MM-DD-YYYY")}&end_date=${moment(this.state.end_date).format("MM-DD-YYYY")}&period=${this.state.selected_period}`
+		})
```
taimur38: do we want to push to history? I would think that we would want to replace the state without leaving a trace of all these state changes. otherwise pressing back will be very annoying

taimur38: there is a way to set all these things in the url, but without the history being appended to. I think its replace state or something

web/src/modules/Analytics/Teacher-Attendance/index.tsx
```diff
@@ -23,33 +24,33 @@ interface Filter {
...
-					.map(([month, { PRESENT, LEAVE, ABSENT }]) => {
-							
-						const percent = ( ABSENT / (ABSENT + PRESENT + LEAVE) * 100)
-							return { month, PRESENT, LEAVE, ABSENT, percent: isFinite(percent) ? percent : 1 }
-						})}>
+					<LineChart data={Object.entries(attendance)
+						.sort(([d1, ], [d2, ]) => moment(d1, date_format).diff(moment(d2, date_format)))
+						.map(([month, { PRESENT, LEAVE, ABSENT }]) => ({
+							month, PRESENT, LEAVE, ABSENT, 
+							percent: ( ABSENT / (ABSENT + PRESENT + LEAVE) * 100).toFixed(2)
```
taimur38: is this calculation still correct with the different kinds of LEAVE we have?

web/src/modules/Analytics/Teacher-Attendance/index.tsx
```diff
@@ -149,17 +169,53 @@ class TeacherAttendanceAnalytics extends Component < propTypes, S > {
...
 		}
 	} 
+
+	onStateChange = () => {
+
+		this.props.history.push({
+			pathname: '/analytics/teacher-attendance',
+			search: `?start_date=${moment(this.state.start_date).format("MM-DD-YYYY")}&end_date=${moment(this.state.end_date).format("MM-DD-YYYY")}&period=${this.state.selected_period}`
+		})
+	}
```
taimur38: same comment on preserving history

web/src/modules/Analytics/Teacher-Attendance/index.tsx
Mudassar045: Yes, it will work fine

web/src/modules/Analytics/Expenses/index.tsx
Mudassar045: I got your points. We shouldn't push to history
Here's code snippet which serving the purpose

> window.history.replaceState(this.state, "Fees Analytics", `${url}?${params}`)

**Issue with above snippet**: 
   When pressing `back button`, routing to `/landing url` in all

PR #176, opened 2019-10-31
setuped trial countdown

raoaliahmad:


web/src/actions/index.ts
```diff
@@ -796,4 +796,21 @@ export const issueCertificate = (type: string, student_id: string, faculty_id: s
 			date
 		}
 	}]))
+}
+
+export const resetTrial = () => (dispatch: Function) => {
+	const date = moment.now()
```
taimur38: we can just use `new Date().getTime()` if we don't need the date formatting.

web/src/modules/Landing/index.js
```diff
@@ -47,13 +47,44 @@ class Landing extends Component {
...
+			return true
+		}
+		else if (word === `${this.props.school_id}${moment(date).format("MMDDYYYY")}`) {
+			//this.props.markPurchased()
+			return true
+		}
+		else {
+			return false
+		}
+	}
```
taimur38: dont need the else if you are using returns inside the if
```javascript
if(cond1) {
   return true
}

if(cond2) {
    return true
}

return false
```

taimur38: Also here is where you should use utils.hash to compare. hashing is async so will need to use promises up and down the logic

web/src/modules/Landing/index.js
```diff
@@ -47,13 +47,44 @@ class Landing extends Component {
...
 	componentDidMount() {
+
+		const { paid, date } = this.props.package_info
+		
 		const container = document.querySelector(".landing .horizontal-scroll-container");
 
 		container.onscroll = () => this.setState({ scroll: container.scrollLeft })
 		container.scrollTo(window.innerWidth, 0)
+		
+		if ( !paid && (15 - moment().diff(date, "days")) > 16) {
```
taimur38: what is this supposed to do?

15 - difference > 16 ??  so if difference is -1?

taimur38: this should read more clearly. and the trial length should be a variable set up top instead of hardcoded

web/src/modules/Landing/index.js
```diff
@@ -67,10 +98,27 @@ class Landing extends Component {
 	getDailyStatsRoute = (stats_type) => {
 		return `/analytics/${stats_type}?start_date=${moment().format('MM-DD-YYYY')}&end_date=${moment().format('MM-DD-YYYY')}&period=Daily`
 	}
+	askForPassword = () => {
+		
+		const magic_word = window.prompt("Please Enter Purchase or Reset Code.", "")
+
+		if (this.checkMagicWord(magic_word)) {
+			console.log("ACCEPTED")
```
taimur38: alert?

web/src/modules/Landing/index.js
```diff
@@ -342,13 +396,16 @@ class Landing extends Component {
...
-	dispatch => ({
-		logout: () => dispatch(createLogout())
-	}))(Landing)
+	user: state.db.faculty[state.auth.faculty_id],
+	students: state.db.students,
+	faculty: state.db.faculty,
+	permissions: state.db.settings.permissions,
+	lastSnapshot: state.lastSnapshot,
+	unsyncd: Object.keys(state.queued).length,
+	package_info: state.db.package_info,
```
taimur38: this will be null so make sure we account for it

taimur38: if theres ever any errors on landing page it makes the entire app unusable so triple check all cases

web/src/modules/Landing/style.css
```diff
@@ -16,6 +16,19 @@
...
+	.trial-bar {
+		background: #fc6171;
+		padding: 10px;
+		text-align: center;
+		width: 100%;
+		color: #fafafa;
+		font-size: 2.2em;
+		transition: all .2s ease-in;
+		cursor: pointer;
+	}
```
taimur38: are we sure this doesn't mess up any of the other calculations? like if you have a small screen and have to scroll down?

web/src/utils/indexedDb.ts
```diff
@@ -32,6 +32,10 @@ export const initState : RootReducerState = {
 			schoolLogo:""
 		},
 		max_limit: -1,
+		package_info: {
+			date: moment().subtract(13,"days").unix() * 1000, //THIS IS THE PROBLEM
```
taimur38: strange....

double check if this is getting overwritten when we spread over the init state later. it probably is

web/src/utils/localStorage.ts
```diff
@@ -29,6 +30,10 @@ const initState : RootReducerState = {
 		assets:{
 			schoolLogo:""
 		},
+		package_info: {
+			date: moment().subtract(3,"days").unix() * 1000,
+			paid: false
+		},
```
taimur38: also double check what happens when we spread over this, if it's overwritten

elixir/lib/sarkar/auth/auth.ex
```diff
@@ -10,6 +10,22 @@ defmodule Sarkar.Auth do
...
+				"value" => %{
+					"paid" => false,
+					"trial_period" => 15,
+					"date" => :os.system_time(:millisecond)
+				},
+				"path" => ["db","package_info"],
+				"type" => "MERGE",
+				"client_id" => "backend"
+			}
+		})
```
taimur38: Am not sure about using save here. Should be the same as what you used in the task, with sync.

Also, this logic should be inside `school.ex`, not inside auth. Should just call a function like `Sarkar.School.init_trial(args)` and that will call sync_changes etc with all the details

PR #177, opened 2019-10-31
Printing work - WIP

Mudassar045:
Add new print components for the following modules:
- [x] All students
- [x] Outstanding fee list in Fee Analytics
- [x] Expense Module (General Expense and Income-Expenditure)

 **NOTE**: There are some `CSS Styles` rules overriding issues, will resolve them once you review `PR`

**Preview Links**:
- [Student List](https://www.dropbox.com/s/6vp6vfc2kdn4zpc/studentlist.pdf?dl=0)
- [Outstanding Fee List](https://www.dropbox.com/s/a14000tiwlnwhbd/oustfeelist.pdf?dl=0)
- [Expense List](https://www.dropbox.com/s/st1gyg2vehtfyzg/Expens-list.pdf?dl=0)
- [Income-Expenditure List](https://www.dropbox.com/s/km7mbq8sdyw53ih/income-expenditure.pdf?dl=0)


web/src/components/Printable/Expense/General/style.css
```diff
@@ -0,0 +1,24 @@
...
+    }
+    .row-amount {
+        width: 15% !important;
+    }
+    .row-category {
+        width: 20% !important;
+    }
+    .row-label {
+        width: 25% !important;
+    }
```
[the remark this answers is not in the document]
taimur38: ah now I see what you're doing.

You're basically hardcoding the widths for each print table column. And since in each you are using these same classes and selection rules, you have to use the important tag. But these will all overwrite each other. When we import a style sheet for a react component, it doesn't only apply to that component - it is put into a global css scope and applies to the entire site. For example, `.row-quantity` here is at 10%, and further down you have it as 15%. Since you're using the same selection rules, only one will apply in both cases, and here the `!important` one will win out.

I understand you want to make certain columns skinnier than other columns. I think the best way to do this is to use `flex-grow` and `flex-shrink` properties. Then you can specify that a column should shrink more than the other one if it needs to fit in. 

Or you need to come up with actual generic columns that apply everywhere. In effect, thats what you've done here but there is a lot of extra css generated. You need to just have one set of rules instead of one per component.

web/src/modules/Analytics/Fees/index.tsx
```diff
@@ -3,12 +3,13 @@ import { Link, RouteComponentProps } from 'react-router-dom'
...
-import { PrintHeader } from 'components/Layout'
-import Former from 'utils/former'
+import { addMultiplePayments } from '../../../actions'
+import Former from '../../../utils/former'
 import checkDuesAsync from 'utils/calculateDuesAsync'
-import { numberWithCommas } from 'utils/numberWithCommas'
-import { getSectionsFromClasses } from 'utils/getSectionsFromClasses'
+import { numberWithCommas } from '../../../utils/numberWithCommas'
+import { getSectionsFromClasses } from '../../../utils/getSectionsFromClasses'
+import { chunkify } from '../../../utils/chunkify'
```
taimur38: don't need these relative imports, can just write `from 'utils/chunkify'` etc

web/src/modules/Student/List/index.js
```diff
@@ -1,17 +1,16 @@
 import React, {Component} from 'react';
 import { connect } from 'react-redux';
 import { Link } from 'react-router-dom';
-import qs from 'query-string'
+import qs from 'querystring'
```
taimur38: the dependency changed?

web/src/utils/chunkify.ts
```diff
@@ -0,0 +1,18 @@
...
+            if(!agg[chunkIndex]) {
+              agg[chunkIndex] = [] // create a new chunk
+            }
+            agg[chunkIndex].push(curr)
+            return agg
+        }, [])
+    }
+}
+
+export default chunkify
```
taimur38: nice. I would just add a comment section above the function explaining what it does - takes a single array, breaks it into chunks of size chunkSize, and returns an array of arrays of max length chunkSize.

taimur38: also can clean the code a bit.
```javascript
if(arr.length === 0) {
    return [[]]
}

if (arr.length <= chunkSize) {
    return [kvArray]
}

return arr.reduce....
```

basically you dont need the nested else statements because the `return` inside the if statement will jump you out of the circuit

web/src/modules/Student/List/index.js
Mudassar045: No, both `query-string` and `querystring` are available dependencies.  When we convert to typescript, unable to find this (`query-string`) dependency. In some old `.js` files we're using 'query-string', and now when I converted some files to `ts` this (`query-string`) causing the error.

web/src/utils/chunkify.ts
Mudassar045: I will do that. Yes It would be nice to have method explanation

web/src/modules/Analytics/Fees/index.tsx
Mudassar045: It was old files, before updating the dependencies. Will fix it.

web/src/components/Printable/Expense/Other/list.tsx
```diff
@@ -0,0 +1,45 @@
+import React from "react"
+import "../../print.css"
```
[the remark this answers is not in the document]
Mudassar045: Yes, I encountered this problem before, I just want to show you the demo. will make a generic solution for the `css styles`.

web/src/components/Printable/Expense/General/style.css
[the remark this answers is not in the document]
Mudassar045: Yes, I got it. I will make actual generic solution for the columns and also for the `CSS styles` files.

PR #186, opened 2019-11-07
show result cards by sections

Mudassar045:
It's a dependency task for print result card for the section on a single sheet, so it needs to be done before that task.

### DIFF
--- a/web/src/modules/Class/Single/ReportsMenu/index.js
+++ b/web/src/modules/Class/Single/ReportsMenu/index.js
@@ -5,6 +5,7 @@ import Former from 'utils/former'
 import { StudentMarks, reportStringForStudent } from 'modules/Student/Single/Marks'
 import { smsIntentLink } from 'utils/intent'
 import { logSms } from 'actions'
+import { getSectionsFromClasses } from 'utils/getSectionsFromClasses'
 
 
 import './style.css'
@@ -28,7 +29,7 @@ class ClassReportMenu extends Component {
 	}
 
 	logSms = (messages) => {
-		if(messages.length === 0){
+		if (messages.length === 0) {
 			console.log("No Message to Log")
 			return
 		}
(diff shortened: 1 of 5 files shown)


web/src/modules/Class/Single/ReportsMenu/index.js
```diff
@@ -38,26 +39,26 @@ class ClassReportMenu extends Component {
...
 		const subjects = new Set()
 		const examSet = new Set()
-		
-		for(const s of relevant_students)
-		{
-			for(const e of Object.values(exams))
-			{ 
-				if(e.section_id === s.section_id){
+
+		for (const s of relevant_students) {
```
taimur38: shouldn't this be a `let` inside a loop? since it is reassigned per iteration

web/src/modules/Student/Single/Marks/index.js
```diff
@@ -41,7 +42,10 @@ class StudentMarksContainer extends Component {
 		const {match, students, settings, sms_templates, exams, classes } = this.props;
 		const id = match.params.id;
 		const student = students[id];
-		const curr_class = Object.values(classes).find(c => c.sections[student.section_id]!== undefined)		
+
+		const curr_section = getSectionsFromClasses(classes)
+			.filter( section  => student.section_id !==undefined && student.section_id === section.id)[0]
```
taimur38: the linter doesn't complain about this spacing?

web/src/modules/Class/Single/ReportsMenu/index.js
Mudassar045: We can use `let` instead of `const` too, if we reassign the variable inside the block. but we're not reassigning the variable `s` (student object) in the `for...of` block.

web/src/modules/Student/Single/Marks/index.js
Mudassar045: No!

taimur38: It should

Mudassar045: I'll check again in a while.

Mudassar045: No, it doesn't complain about the spacing

PR #197, opened 2019-12-03
Track app

raoaliahmad:
Should look at it and let me know what do you think

### DIFF
--- a/elixir/lib/sarkar/action_handlers/mis.ex
+++ b/elixir/lib/sarkar/action_handlers/mis.ex
@@ -53,41 +53,43 @@ defmodule Sarkar.ActionHandler.Mis do
 		{:reply, succeed(), state}
 	end
 
-	def handle_action(%{"type" => "SYNC", "payload" => payload, "lastSnapshot" => last_sync_date}, %{school_id: school_id, client_id: client_id} = state) do
+	def handle_action(
+		%{
+			"type" => "SYNC",
+			"payload" => %{"analytics" => analytics, "mutations" => mutations },
+			"lastSnapshot" => last_sync_date
+		},
+		%{
+			school_id: school_id,
+			client_id: client_id
+		} = state
+	) do
 
-		old_syncs = payload 
-			|> Map.drop(["ANALYTICS","MUTATION"])
-
-		old_res = if map_size(old_syncs || %{}) > 0 do
-			IO.puts "Got OLD SYNCS from #{school_id}"
-			Sarkar.School.sync_changes(school_id, client_id, old_syncs, last_sync_date)
-		end
-
-		mutations = payload
-		|> Map.get("MUTATION")
-
+		IO.puts "RuNNING NEW SYNC"
 		mutation_res = if map_size(mutations || %{}) > 0 do
-			IO.puts "Got mutations SYNCS"
-			res = Sarkar.School.sync_changes(school_id, client_id, mutations, last_sync_date)
+			Sarkar.School.sync_changes(school_id, client_id, mutations, last_sync_date)
 		end
 
-		analytics = payload
-		|> Map.get("ANALYTICS")
-
-		analytics_res = if map_size(analytics || %{}) > 0 do
-			IO.puts "Got analytics SYNCS"
-			res = Sarkar.Analytics.add_writes(school_id, client_id, analytics, last_sync_date)
-		end
+		analytics_res = Sarkar.Analytics.record(school_id, client_id, analytics, last_sync_date)
 
 		res = %{
-			"MUTATION" => mutation_res,
-			"ANALYTICS" => analytics_res,
-			"OLD" => old_res
+			"mutations" => mutation_res,
+			"analytics" => analytics_res,
 		}
 
 		IO.inspect res
-		
 		{:reply, succeed(res), state}
+	end
+
+	def handle_action(%{"type" => "SYNC", "payload" => payload, "lastSnapshot" => last_sync_date}, %{school_id: school_id, client_id: client_id} = state) do
+
+		IO.puts "Got OLD SYNCS from #{school_id}"
+		IO.inspect payload
+
+		res = Sarkar.School.sync_changes(school_id, client_id, payload, last_sync_date)
+
+		{:reply, succeed(res), state}
+
 	end
 
 	def handle_action(%{"type" => "SMS", "payload" => payload}, %{school_id: school_id, client_id: client_id} = state) do
--- a/elixir/lib/sarkar/action_handlers/mis.ex
+++ b/elixir/lib/sarkar/action_handlers/mis.ex
@@ -54,7 +54,39 @@ defmodule Sarkar.ActionHandler.Mis do
 	end
 
 	def handle_action(%{"type" => "SYNC", "payload" => payload, "lastSnapshot" => last_sync_date}, %{school_id: school_id, client_id: client_id} = state) do
-		res = Sarkar.School.sync_changes(school_id, client_id, payload, last_sync_date)
+
+		old_syncs = payload 
+			|> Map.drop(["ANALYTICS","MUTATION"])
+
+		old_res = if map_size(old_syncs || %{}) > 0 do
+			IO.puts "Got OLD SYNCS from #{school_id}"
+			Sarkar.School.sync_changes(school_id, client_id, old_syncs, last_sync_date)
+		end
+
+		mutations = payload
+		|> Map.get("MUTATION")
+
+		mutation_res = if map_size(mutations || %{}) > 0 do
+			IO.puts "Got mutations SYNCS"
+			res = Sarkar.School.sync_changes(school_id, client_id, mutations, last_sync_date)
+		end
+
+		analytics = payload
+		|> Map.get("ANALYTICS")
+
+		analytics_res = if map_size(analytics || %{}) > 0 do
+			IO.puts "Got analytics SYNCS"
+			res = Sarkar.Analytics.add_writes(school_id, client_id, analytics, last_sync_date)
+		end
+
+		res = %{
+			"MUTATION" => mutation_res,
+			"ANALYTICS" => analytics_res,
+			"OLD" => old_res
+		}
+
+		IO.inspect res
+		
 		{:reply, succeed(res), state}
 	end
 
--- a/elixir/lib/sarkar/action_handlers/mis.ex
+++ b/elixir/lib/sarkar/action_handlers/mis.ex
@@ -65,7 +65,6 @@ defmodule Sarkar.ActionHandler.Mis do
 		} = state
 	) do
 
-		IO.puts "RuNNING NEW SYNC"
 		mutation_res = if map_size(mutations || %{}) > 0 do
 			Sarkar.School.sync_changes(school_id, client_id, mutations, last_sync_date)
 		end
@@ -77,14 +76,10 @@ defmodule Sarkar.ActionHandler.Mis do
 			"analytics" => analytics_res,
 		}
 
-		IO.inspect res
 		{:reply, succeed(res), state}
 	end
 
 	def handle_action(%{"type" => "SYNC", "payload" => payload, "lastSnapshot" => last_sync_date}, %{school_id: school_id, client_id: client_id} = state) do
-
-		IO.puts "Got OLD SYNCS from #{school_id}"
-		IO.inspect payload
 
 		res = Sarkar.School.sync_changes(school_id, client_id, payload, last_sync_date)
 
--- a/elixir/lib/sarkar/action_handlers/mis.ex
+++ b/elixir/lib/sarkar/action_handlers/mis.ex
@@ -108,10 +108,10 @@ defmodule Sarkar.ActionHandler.Mis do
 		{:reply, fail("Please update your mischool app to the latest version."), state}
 	end
 
-	def handle_action(%{"type" => type, "payload" => payload}, state) do
+	def handle_action(%{"type" => type, "payload" => payload} = msg, state) do
 		IO.puts "it is likely you have not authenticated. no handler exists for this combination of state and message"
 		IO.inspect type
-		IO.inspect payload
+		IO.inspect msg 
 		IO.inspect state
 		{:reply, fail("Please update your mischool app to the latest version."), state}
 	end
--- a/elixir/lib/sarkar/analytics/analytics.ex
+++ b/elixir/lib/sarkar/analytics/analytics.ex
@@ -1,0 +1,68 @@
+
+defmodule Sarkar.Analytics do
+
+	def add_writes(school_id, client_id, changes, last_sync_date) do
+
+		{new_writes, last_date} = changes
+		|> Enum.sort(fn({ _, %{"date" => d1}}, {_, %{"date" => d2}}) -> d1 < d2 end)
+		|> Enum.reduce(
+			{ %{}, 0 },
+			fn({ path_key, payload}, {agg_new_writes, max_date}) ->
+				
+				%{
+					"action" => %{
+						"path" => path,
+						"type" => type,
+						"value" => value
+					},
+					"date" => date
+				} = payload
+
+				p_key = Enum.join(path, ",")
+
+				write = %{
+					"date" => date,
+					"value" => value,
+					"path" => path,
+					"type" => type,
+					"client_id" => client_id
+				}
+
+				case Map.get(agg_new_writes, p_key) do
+					nil -> 
+						{
+							Map.put(agg_new_writes, p_key, write),
+							max(date, max_date)
+						}
+					%{"date" => prev_date, "value" => prev_value} when prev_date < date ->
+						{
+							Map.put(agg_new_writes, p_key, write),
+							max(date, max_date)
+						}
+					%{"date" => prev_date, "value" => prev_value} when prev_date >= date ->
+						# IO.puts "#{school_id}: #{prev_date} is more recent than #{date}. current time is #{:os.system_time(:millisecond)}"
+						# IO.inspect write
+						{
+							agg_new_writes,
+							max_date
+						}
+					other -> 
+						IO.puts "OTHER!!!!!!!!!!!!!"
+						IO.inspect other
+						{
+							Map.put(agg_new_writes, p_key, write),
+							max(date, max_date)
+						}
+				end
+			end
+		)
+
+		case Sarkar.Store.School.save_writes(school_id,new_writes) do
+			{:ok} ->
+				%{"type" => "CONFIRM_ANALYTICS_SYNC"}
+			{:err, err}->
+				IO.inspect err
+				{:err}
+		end
+	end
+end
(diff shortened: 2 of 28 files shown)


web/src/actions/index.ts
```diff
@@ -824,4 +824,17 @@ export const markPurchased = () => (dispatch: Function) => {
 		path: ["db", "package_info", "paid"],
 		value: true
 	}]))
+}
+
+export const trackRoute = (route: string) => (dispatch: Function) => {
+	const time = moment.now()
+
```
taimur38:
```javascript

dispatch(analyticsEvent({
   type: "ROUTE",
   meta: {
      route
   }
}
```

web/src/modules/Landing/index.js
```diff
@@ -430,7 +430,7 @@ export default connect(state => ({
 	faculty: state.db.faculty,
 	permissions: state.db.settings.permissions,
 	lastSnapshot: state.lastSnapshot,
-	unsyncd: Object.keys(state.queued).length,
+	unsyncd: Object.keys(state.queued["MUTATION"] || {}).length + Object.keys(state.queued["ANALYTICS"] || {}).length,
```
taimur38: should only check mutations

web/src/reducers/index.js
```diff
@@ -7,6 +7,18 @@ const rootReducer = (state, action) => {
...
 	switch (action.type) {
 
+		case "CONFIRM_ANALYTICS_SYNC": {
+			return {
+				...state,
+				queued: {
+					...state.queued,
+					"ANALYTICS": {}
+				},
+				acceptSnapshot: true,
```
taimur38: shouldn't affect acceptSnapshot

taimur38: and dont update the lastSnapshot

web/src/typings.d.ts
```diff
@@ -45,14 +45,26 @@ interface RootReducerState {
...
+		},
+		"ANALYTICS": {
+			[path: string]: {
+				action: {
+					path: string[];
+					value?: any;
+					type: "ANALYTICS_MERGE"
+				};
+				date: number
+			}
```
taimur38: analytics is append only and probably shouldn't be an object. can be an array of analytics event types. and there should be a definition for the type of analytics as per my comment above

elixir/lib/sarkar/action_handlers/mis.ex
```diff
@@ -53,43 +53,45 @@ defmodule Sarkar.ActionHandler.Mis do
...
 
 		IO.inspect res
-		
 		{:reply, succeed(res), state}
 	end
 
+	def handle_action(%{"type" => "SYNC", "payload" => payload, "lastSnapshot" => last_sync_date}, %{school_id: school_id, client_id: client_id} = state) do
+
+		IO.puts "Got OLD SYNCS from #{school_id}"
+		IO.inspect payload
```
taimur38: don't need the inspects

elixir/lib/sarkar/analytics/analytics.ex
```diff
@@ -1,68 +1,24 @@
 
 defmodule Sarkar.Analytics do
 
-	def add_writes(school_id, client_id, changes, last_sync_date) do
+	def record(school_id, client_id, changes, last_sync_date) do
```
taimur38: they arent changes, they are events

elixir/lib/sarkar/analytics/analytics.ex
```diff
@@ -1,68 +1,24 @@
...
+		res = changes
+			|> Enum.each(
+				fn (%{"time" => time, "value" => %{"meta" => meta, "type" => type}}) ->
```
taimur38: "type" probably shouldn't be inside a random "value" object.. whats the point of structuring it this way?

elixir/lib/sarkar/analytics/analytics.ex
```diff
@@ -1,68 +1,24 @@
...
-							max(date, max_date)
-						}
+					case Postgrex.query(Sarkar.School.DB,
+						"INSERT INTO mischool_analytics ( school_id, value, time, type, client_id) VALUES ($1, $2, $3, $4, $5)",
+					[school_id, meta, time, type, client_id]) do
+					{:ok, resp} ->
+						{:ok, resp}
+					{:error, err} -> 
+						IO.puts "ERROR PUTTING IN DB"
+						IO.inspect err
```
taimur38: should return something

taimur38: also should not run a bunch of inserts as separate queries. batch these together into one insert

elixir/lib/sarkar/analytics/analytics.ex
```diff
@@ -1,68 +1,24 @@
...
+		IO.inspect res
 
-		case Sarkar.Store.School.save_writes(school_id,new_writes) do
-			{:ok} ->
-				%{"type" => "CONFIRM_ANALYTICS_SYNC"}
-			{:err, err}->
-				IO.inspect err
-				{:err}
-		end
+		%{"type" => "CONFIRM_ANALYTICS_SYNC"}
```
taimur38: should also return the latest time associated with this batch of events. that way if anything happened in between on the client, it can only wipe out the relevant ones from the queue instead of everything

web/src/actions/core.ts
```diff
@@ -18,43 +18,51 @@ export interface MergeAction {
...
-		[curr.path.join(',')]: {
-			action: {
-				type: "ANALYTICS_MERGE",
-				path: curr.path.map(p => p === undefined ? "" : p),
-				value: curr.value
-			},
-			date: new Date().getTime()
-		}
-	}), {})
+export const analyticsEvent = (event: AnalyticsEvent[]) => (dispatch: (a: any) => any, getState: () => RootReducerState, syncr: Syncr) => {
```
taimur38: think dispatch can just use type `Dispatch` now...

web/src/actions/core.ts
```diff
@@ -18,43 +18,51 @@ export interface MergeAction {
...
+				},
+				time: new Date().getTime()
+			}]
+	}, [])
 
+	
 	const state = getState();
-
-	const rationalized_merges = {
+	const rationalized_event_payoad = {
```
taimur38: payload not payoad

web/src/actions/core.ts
```diff
@@ -18,43 +18,51 @@ export interface MergeAction {
...
+		return [
+			...agg,
+			{
+				type: "ANALYTICS_EVENT",
+				value: {
+					type: curr.type,
+					meta: curr.meta
+				},
+				time: new Date().getTime()
+			}]
```
taimur38: why are we shaping the object this way? there doesn't seem to be any benefit of `type: "ANALYTICS_EVENT"`.

why not
```javascript
{
   type: curr.type,
   meta: curr.meta,
   time: new Date().getTime()
}
```

web/src/actions/core.ts
```diff
@@ -217,8 +222,8 @@ export const createDeletes = (paths: Delete[]) => (dispatch: Dispatch<AnyAction>
 	
 	const rationalized_deletes = {
 		...state.queued,
```
taimur38: pretty sure you want ...state.queued.mutations

also theres no point in using the keys like this. if there is a benefit to using keys, it is going to be declaring them at the top of the file as a const once. this way actually just makes it harder to read but without the benefit of being able to change them if needed. so i would just remove all mention of key and use mutation / analytics

web/src/actions/core.ts
```diff
@@ -274,10 +279,10 @@ export interface SnapshotDiffAction {
...
 // queue up an object where key is path, value is action/date
-interface Queuable {
+interface Queuable { 
 	[path: string]: {
 		action: {
-			type: "MERGE" | "DELETE" | "ANALYTICS_MERGE";
+			type: "MERGE" | "DELETE" | "ANALYTICS_EVENT";
 			path: string[];
 			value?: any;
 		};
```
taimur38: This type needs to change. If this is why you were structuring the events the way you were. 

there are now two types of queuable - the kind that goes into mutation, and the kind that goes into analytics. There are a few different ways to approach this. One is you can fire different actions types and have 2 reducers that are modifying the same part of the tree. the other is you can create a base type which is differentiated on another type variable that tells it either "mutations" or "analytics". you basically have an abstract type and then the 2 types which inherit from it. For this you can again look at how we are handling call events in the supplier portal typings

taimur38: but you shouldn't be wrangling the types to fit into the previous object shape

web/src/actions/core.ts
```diff
@@ -327,20 +340,21 @@ export const connected = () => (dispatch: (a: any) => any, getState: () => RootR
...
-			})
+			.then(resp => triActionCaller(resp,dispatch))		
 			.catch(err => {
 				console.error(err)
 				alert("Authorization Failed. Log out and Log in again.")
 			})
 	}
 }
 
+export const triActionCaller = (resp: { key: string, val: any },dispatch: Dispatch) => {
```
taimur38: should call this something like `multiAction`, and instead of passing the dispatch function along, you should just use it as an async action like the rest of these.

`.then(res => dispatch(multiAction(res)))`

web/src/reducers/index.js
```diff
@@ -12,10 +12,8 @@ const rootReducer = (state, action) => {
 				...state,
 				queued: {
 					...state.queued,
-					"ANALYTICS": {}
-				},
-				acceptSnapshot: true,
-				lastSnapshot: new Date().getTime()
+					analytics: []
+				}
```
taimur38: confirm analytics sync should come back with a timestamp of the latest analytics event it processed. then you should only delete the items from analytics which are less than that time

web/src/typings.d.ts
```diff
@@ -55,16 +55,14 @@ interface RootReducerState {
...
-		"ANALYTICS": {
-			[path: string]: {
-				action: {
-					path: string[];
-					value?: any;
-					type: "ANALYTICS_MERGE"
-				};
-				date: number
+		"analytics": Array<{
+			type: "ANALYTICS_EVENT",
```
taimur38: its just not useful to have this variable there

web/src/typings.d.ts
```diff
@@ -55,16 +55,14 @@ interface RootReducerState {
...
+			value: {
+				type: "ROUTE"
```
taimur38: and this probably shouldn't be hardcoded here

web/src/utils/indexedDb.ts
```diff
@@ -12,8 +12,8 @@ const defaultTemplates = () => ({
 export const initState: RootReducerState = {
 	client_id: localStorage.getItem("client_id") || v4(),
 	queued: {
-		"MUTATION": {},
-		"ANALYTICS": {}
+		"mutations": {},
```
taimur38: dont need quotation marks around it

web/src/utils/indexedDb.ts
```diff
@@ -106,8 +106,33 @@ export const loadDb = async () => {
 			}
 		}
 
-		const prev: RootReducerState = JSON.parse(serialized)
+		let prev: RootReducerState = JSON.parse(serialized)
 		const client_id = localStorage.getItem('client_id') || prev.client_id || v4()
+
+		if ((!prev.queued.mutations || !prev.queued.analytics)) {
```
taimur38: why double parenthesis?

web/src/utils/indexedDb.ts
```diff
@@ -106,8 +106,33 @@ export const loadDb = async () => {
...
+						...prev.queued
+					}
+				}
+			}
+			else {
+				prev.queued = {
+					analytics: [],
+					mutations: {}
+				}
+			}
```
taimur38: actually dont need this if statement. you can call spread on an empty object, just make sure that prev.queued is not undefined

web/src/utils/localStorage.ts
```diff
@@ -13,8 +13,8 @@ const initState: RootReducerState = {
 	client_id: v4(),
 	initialized: false,
 	queued: { 
-		"MUTATION": {},
-		"ANALYTICS": {}
+		"mutations": {},
+		"analytics": []
```
taimur38: dont need quotations

elixir/lib/sarkar/analytics/analytics.ex
raoaliahmad: @taimur38 thats because there were two types before namalytics_merge and a sub tyoe i.e route, sms. but now that i removed the analytics_merge I don't need this.

elixir/lib/sarkar/analytics/analytics.ex
```diff
@@ -1,24 +1,69 @@
...
 
-					case Postgrex.query(Sarkar.School.DB,
-						"INSERT INTO mischool_analytics ( school_id, value, time, type, client_id) VALUES ($1, $2, $3, $4, $5)",
-					[school_id, meta, time, type, client_id]) do
-					{:ok, resp} ->
-						{:ok, resp}
-					{:error, err} -> 
-						IO.puts "ERROR PUTTING IN DB"
-						IO.inspect err
+		if map_size(events) > 0 do
```
taimur38: instead of a big if statement at the top here, you should use pattern matching to overload the record function in the case of events being of size 0. there is an example of this here https://github.com/CERP/edmarkaz/blob/master/elixir/lib/sarkar/store/supplier.ex#L101-L118

web/src/actions/core.ts
```diff
@@ -23,45 +24,45 @@ export interface AnalyticsEvent {
...
-		payload: rationalized_event_payoad
+		payload: rationalized_event_payload
 	}
 
 	syncr.send(payload)
-		.then(res => triActionCaller(res, dispatch))
+		.then(res => {
+			dispatch(QueueAnalytics(event_payload, "analytics"))
+			dispatch(multiAction(res))
+		})
```
taimur38: this doesn't make sense. if the sync is successful then we are queueing analytics?

I thought we would queue before we send no matter what

web/src/actions/core.ts
```diff
@@ -23,45 +24,45 @@ export interface AnalyticsEvent {
...
 		.catch(err => {
-			dispatch(QueueAnalytics(event_payload, key))
-			
+			if (!state.connected) {
+				dispatch(QueueAnalytics(event_payload, "analytics"))
+			}
```
taimur38: I don't think you need an if statement here

web/src/actions/core.ts
```diff
@@ -110,11 +111,14 @@ export const createMerges= (merges: Merge[]) => (dispatch: (a: any) => any, getS
 	}
 
 	syncr.send(payload)
-		.then(res => triActionCaller(res,dispatch))
-		.catch(err => {
-
+		.then(res => {
 			dispatch(QueueUp(new_merges, key))
```
taimur38: again this doesnt make sense. you need to queue up BEFORE you send, not after you receive a successful sync

web/src/actions/core.ts
taimur38: you haven't actually changed the type of the queueable or queue to account for analytics. you just made a separate queueanalytics event which takes action type any...

web/src/actions/core.ts
```diff
@@ -415,3 +416,10 @@ export const loadDB = () => (dispatch: Function, getState: () => RootReducerStat
 
 	})
 }
+
+export const multiAction = (resp: { key: string, val: any }) => {
+	for (let [key, action] of Object.entries(resp)) {
+		if(action)
+			return (action)
```
taimur38: this loops and just returns the first action? it's not dispatching all actions in the resp

web/src/typings.d.ts
```diff
@@ -41,11 +41,17 @@ interface RootDBState {
 	diary : MISDiary
 }
 
+interface AnalyticsQueue {
+	type: string;
+	meta: any;
+	time: number;
+}
```
taimur38: not a queue - this is a base analytics event.

web/src/actions/core.ts
```diff
@@ -24,7 +24,7 @@ export interface AnalyticsEvent {
 	meta: any
 }
 
-export const analyticsEvent = (event: AnalyticsEvent[]) => (dispatch: Dispatch, getState: () => RootReducerState, syncr: Syncr) => {
+export const analyticsEvent = (event: AnalyticsEvent[]) => (dispatch: Function, getState: () => RootReducerState, syncr: Syncr) => {
```
taimur38: Why did this change from Dispatch to Function?

web/src/actions/core.ts
```diff
@@ -298,24 +296,40 @@ interface Queuable {
...
+	queue_type: "mutations"
 }
 
-export const QueueUp = (action: Queuable, type: "mutations" ) => {
+export interface QueueAnalyticsAction {
+	type: "QUEUE";
+	payload: { [id: string]: BaseAnalyticsEvent }
+	queue_type: "mutations"
+}
+
```
taimur38: so there should be a base queue action, then analytics and mutation queue actions would extend it.

```typescript

interface BaseQueueAction {
   type: "QUEUE"
   queue_type: string
}

export interface QueueAnalyticsAction extends QueueAction {
    queue_type: "analytics"
    payload: { [id: string]: BaseAnalyticsEvent }
}

export interface QueueMutationsAction extends QueueAction {
    queue_type: "mutations"
    payload: MutationQueuable
}

export interface QueueAction = QueueMutationsAction | QueueAnalyticsAction
```

Then in the reducer, the action comes as a QueueAction. and when you do an if statement
```typescript
const queueAction = action as QueueAction

if(queueAction.queue_type === "mutations") { 
    // in here all the autocomplete etc will work because we automatically know it is of type QueueMutationsAction
}
```

web/src/typings.d.ts
```diff
@@ -41,7 +41,7 @@ interface RootDBState {
 	diary : MISDiary
 }
 
-interface AnalyticsQueue {
+interface BaseAnalyticsEvent {
 	type: string;
 	meta: any;
 	time: number;
```
taimur38: and I assume somewhere that there is a definition
```typescript
interface RouteAnalyticsEvent extends BaseAnalyticsEvent {
    type: "ROUTE",
    meta: { route: string }
}
```

web/src/actions/core.ts
taimur38: an example of how we are using this: 

https://github.com/CERP/edmarkaz/blob/master/web/src/typings.d.ts#L21-L53

web/src/actions/core.ts
```diff
@@ -297,27 +292,31 @@ interface MutationQueuable {
...
+}
+export type QueueAction = QueueMutationsAction | QueueAnalyticsAction
+
+export interface ConfirmAnalyticsSyncAction {
+	type: "CONFIRM_ANALYTICS_SYNC";
+	time: number
 }
 
-export const QueueUp = (action: MutationQueuable ) => {
+export const QueueUp = (action: Queuable ) => {
```
taimur38: QueueUp should be QueueMutation and Queuable should be a MutationQueuable

taimur38: and both these queue methods should specify that they are returning AnalyticsAction or mutationAction etc. otherwise when you change the definitions they won't error before compiling

PR #200, opened 2019-12-09
re-format the fee voucher for single student

Mudassar045:
Check out the demos links

[Before - Single fee voucher](https://www.dropbox.com/s/1plaw3tndllap04/mischool-redesign-print-voucher-2.pdf?dl=0)

[After - Single Fee Voucher](https://www.dropbox.com/s/pdo5zm0wnc5bmjh/mischool-redesign-print-voucher.pdf?dl=0)

### DIFF
--- a/web/src/modules/Student/Single/Fees/StudentLedgerPage.tsx
+++ b/web/src/modules/Student/Single/Fees/StudentLedgerPage.tsx
@@ -11,7 +11,7 @@ interface StudentLedgerPageProp {
 	class_name: string
 	settings: RootDBState["settings"]
 	voucherNo: number
-	css_style: string
+	css_style?: "print-only" | "no-print" | ""
 }
 
 export const StudentLedgerPage : React.SFC < StudentLedgerPageProp > = ({ payments, student, settings, class_name, voucherNo, css_style }) => {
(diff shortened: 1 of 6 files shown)


web/src/modules/Student/Single/Fees/StudentLedgerPage.tsx
```diff
@@ -11,44 +11,40 @@ interface StudentLedgerPageProp {
 	class_name: string
 	settings: RootDBState["settings"]
 	voucherNo: number
+	css_style: string
```
taimur38: this should not just be a string, it should be an enum. if i just put random gibberish here, it won't work. so you should give only a few options. secondly if people are having to pass empty strings here then this should be an optional parameter.

Mudassar045: Yes I got it. It probably better to use optional parameter coz I'm passing className to the component.

taimur38: But the other thing to fix here is that css_style shouldn't just be of type string. there are limited values that work so you should express that in the type

css_style: "no-print" | "some-other-option" | "etc"

that way you get autocomplete and the developer has some idea of what the options are instead of having to guess and check

Mudassar045: that's would be much expressive

PR #202, opened 2019-12-11
Family fee voucher

Mudassar045:


web/src/modules/Student/Single/Fees/printPreview.tsx
```diff
@@ -57,45 +50,90 @@ class printPreview extends Component <propTypes, S>{
...
 		}
 
-		return student.payments
+		return this.student().payments
 	}
 
-	render() {
+	student = (): MISStudent => {
+		const id = this.studentID()
+		return this.props.students[id] || this.siblings()[0]
```
taimur38: I think this should be `return id === undefined ? this.siblings()[0] : this.props.students[id]`

in the event that there is somehow an undefined key in the students object

web/src/modules/Student/Single/Fees/printPreview.tsx
```diff
@@ -57,45 +50,90 @@ class printPreview extends Component <propTypes, S>{
...
+	}
 
-		const { classes, student, settings } = this.props
+	siblings = (): MISStudent[] => {
+		const famId = this.familyID()
+		return Object.values(this.props.students)
+			.filter(s => s && s.Name && s.FamilyID && s.FamilyID === famId)
+	}
```
taimur38: can we use memoization here? seems fairly expensive and we call it a decent amount of times.

taimur38: for now it's fine. but will assign you a task for next sprint to look into memoization

web/src/modules/Student/Single/Fees/StudentLedgerPage.tsx
```diff
@@ -74,16 +95,11 @@ export const StudentLedgerPage : React.SFC < StudentLedgerPageProp > = ({ paymen
 				<div className={owed <= 0 ? "advance-amount" : "pending-amount"}><b>Rs. {numberWithCommas(Math.abs(owed))}</b></div>
 			</div>
 			{ // don't show if student ledger rendered in historical fee module, Voucher No. hard coded in this case
-				voucherNo !== 777 && <>
-					<div className="row voucher-signature line">
-						<div>___________________</div>
-						<div>___________________</div>
-					</div>	
+				voucherNo !== 777 &&
```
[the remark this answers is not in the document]
Mudassar045: I used it as default, but will make it optional and then apply check accordingly

web/src/modules/Student/Single/Fees/printPreview.tsx
Mudassar045: I got you.

web/src/modules/Student/Single/Fees/printPreview.tsx
Mudassar045: Yes, we should use memoization here because state changing in the component

PR #214, opened 2020-01-06
add functionality to daily stats to see which student pay the fee

Mudassar045:
- [Print Preview](https://www.dropbox.com/s/srv5wxw184cch9a/daily-ledger.pdf?dl=0)
- Web View:
Web View


web/src/utils/getStudentSection.ts
```diff
@@ -0,0 +1,11 @@
+import getSectionsFromClasses from "./getSectionsFromClasses"
+
+const getStudentSection = (section_id: string, classes: RootDBState["classes"]): AugmentedSection => {
```
taimur38: better name is something like `getSectionFromId`

web/src/modules/Analytics/DailyStats/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+class DailyStats extends Component<PropsType, S> {
+
+    former: Former
+    constructor(props: PropsType) {
+        super(props)
+        
+        const parsed_query = queryString.parse(this.props.location.search)
+
+		const type = parsed_query["?type"] ? parsed_query["?type"].toString() : ''
```
taimur38: Should take 0th element of the list instead of calling `toString()`

web/src/modules/Analytics/DailyStats/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+                    paid_students.push({
+                        ...student,
+                        amount_paid: additional_payment,
+                        section: getStudentSection(student.section_id, classes)
+                    })
+					total_students_paid += 1
+				}
+
+				total_amount_received += additional_payment
+			}
```
taimur38: spacing / alignment here

web/src/modules/Analytics/DailyStats/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+        let total_students_paid = 0
+        let paid_students = [] as AugmentedStudent[]
+
+		for(const student of Object.values(students)) {
+
+			if(student && student.Name) {
+		
+				const additional_payment = Object.values(student.payments || {})
+					.filter(payment => moment(payment.date).format("YYYY-MM-DD") === today_date && payment.type === "SUBMITTED")
+					.reduce((agg, curr) => agg + curr.amount, 0);
```
taimur38: its called `additional_payment` but a better name for it would be something like `amount_paid_today`

web/src/modules/Analytics/DailyStats/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+
+				if(additional_payment > 0) {
+                    paid_students.push({
+                        ...student,
+                        amount_paid: additional_payment,
+                        section: getStudentSection(student.section_id, classes)
```
taimur38: you should include their balance here also, and in the render / print pages as well

web/src/modules/Analytics/DailyStats/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+                        <div className="number">Rs. { total_amount_received }</div>
+                    </div>
+                    <div className="table row student-count">
+                        <label>Total Students Paid: </label>
+                        <div className="number">{ total_students_paid }</div>
+                    </div>
+                    <div className="list">
+                        {
+                            paid_students
+                                .sort((a, b) => a.section.classYear - b.section.classYear)
```
taimur38: you should be able to dual-sort by class year and then enrollment number or name. 

one trick would be like `a.section.classYear * 100 + a.enrollment_number`

web/src/modules/Analytics/DailyStats/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+                <div className="daily-stats">
+                    <div className="title no-print">Daily Statistics</div>
+                    <div className="row date no-print">
+                        <input type="date" 
+                            onChange={this.former.handle(["statsDate"])} 
+                            value={moment(statsDate).format("YYYY-MM-DD")} 
+                            placeholder="Current Date" />
+                    </div>
+                    {
+                        this.renderSection()
```
taimur38: this should be replaced probably with Routes and the tab system we use in analytics and everywhere else. Eventually you'll do something similar for student and teacher attendances, etc.

web/src/modules/Analytics/DailyStats/index.tsx
Mudassar045: cool, will try this

web/src/modules/Analytics/DailyStats/index.tsx
Mudassar045: For now I'm keeping as it, but on Monday will do some replacement on staging, so will integrate tab system

PR #229, opened 2020-02-17
add tag functionality and haslogin field

Mudassar045:
This PR contains:
1- add new property to check to filter login users
2- add tags functionality
3- add tag filter in teachers list


web/src/modules/Teacher/List/index.js
```diff
@@ -27,8 +31,17 @@ export const TeacherList = (props) => {
...
+	Object.values(props.teachers || {})
+		.filter(f => f.id && f.Name)
+		.forEach(s => {
+			Object.keys(s.tags || {})
+				.forEach(tag => tags.add(tag))
+		})
+
 	const items = Object.entries(props.teachers)
-			.filter(([,f]) => f.Name && f.id)
+			.filter(([,f]) => f.Name && f.id && (f.tags && tag !== "" ? Object.keys(f.tags).includes(tag) : true))
```
taimur38: `Object.keys(f.tags).includes(tag)` will go through the "tags" object, convert it into an array, and then do a for loop through the entire array to see if "tag" is in there.

More efficient would be: `f.tags[tag] !== undefined` or maybe even just `f.tags[tag]`. Since we're checking key/value in an object, this is constant time lookup

web/src/modules/Teacher/Single/Create/index.tsx
```diff
@@ -187,7 +190,7 @@ class CreateTeacher extends Component<propTypes, S> {
 		}, 1000);
 	}
 
-	componentWillReceiveProps(newProps: propTypes) {
+	UNSAVE_componentWillReceiveProps(newProps: propTypes) {
```
taimur38: UNSAFE not UNSAVE

web/src/modules/Teacher/Single/Create/index.tsx
```diff
@@ -263,6 +311,20 @@ class CreateTeacher extends Component<propTypes, S> {
...
+					<label>Admin Status</label>
+					<select {...this.former.super_handle(["Admin"])} disabled={!admin}>
+						<option value="true">Admin</option>
+						<option value="false">Not an Admin</option>
+					</select>
+				</div>
+				<div className="row">
+					<label>Has Login</label>
+					<select {...this.former.super_handle(["HasLogin"])} disabled={!admin}>
+						<option value="true">Yes</option>
```
taimur38: I think here the options shouldn't be "Yes" and "No" but more explanatory.

Like "Has login access" and "Does not have login access"

PR #233, opened 2020-03-04
add fees reset  button

Mudassar045:


web/src/actions/index.ts
```diff
@@ -900,4 +900,12 @@ export const removeSubjectFromDatesheet = (id: string, subj: string, section_id:
...
 	}]))
 
+}
+
+export const resetFees = (fees: any) => (dispatch: Function) => {
+	dispatch(createMerges(fees))
+}
+
+export const resetPayments = (payments: any) => (dispatch: Function) => {
+	dispatch(createMerges(payments))
```
taimur38: both of these are completely unclear what they are doing. they are functions that take anything and create merges.

web/src/modules/Student/ManageFees.tsx
```diff
@@ -218,6 +220,46 @@ class ManageFees extends Component<propTypes, S> {
...
+		for (const student of students) {
+			fees.push({
+				path: ["db", "students", student.id, "fees"],
+				value: {}
+			})
+			payments.push({
+				path: ["db", "students", student.id, "payments"],
+				value: {}
+			})
+		}
```
taimur38: Nothing dealing with "path" and "value" should be in these components - this logic needs to be in the action

web/src/modules/Student/Single/style.css
```diff
@@ -5,3 +5,6 @@
 	width: 100%;
 	overflow-x: hidden;
 }
+.section-container {
+	display: block !important;
+}
```
taimur38: you shouldn't need the !important tag. Make the selector more specific if this is the case

web/src/modules/Student/ManageFees.tsx
```diff
@@ -228,73 +270,74 @@ class ManageFees extends Component<propTypes, S> {
...
-								<select {...this.former.super_handle(["selected_student_id"])}>
-									<option value="">Select Student</option>
-									{
-										this.getSelectedSectionStudents().map(s => <option key={s.id} value={s.id}>{s.Name}</option>)
-									}
-								</select>
-							</div> : false}
+				<div className="row" style={{ justifyContent: "flex-end" }}>
+					<div className="button red" onClick={this.resetStudentsFees}>Reset Fees</div>
+				</div>
```
taimur38: This should work the same way as the rest of the module. Where you can remove students fees by class, or for single student, or for all students if you want.

web/src/modules/Student/Single/style.css
Mudassar045: this is to override, the parent child relationship. The below rule is problem.
```css
.layout>div:nth-child(2) {
    display: flex;
    flex-direction: column;
    align-items: center;
    overflow-y: auto;
    padding-bottom: 60px;
}
```
Don't know who wrote this and what's the purpose of this. I know we shouldn't use `!important` tag more frequently but sometime we should. Now I just removed `display: block !important;` and use `inline style` which has more specificity. 
Check this, it's a good article https://wattenberger.com/blog/css-cascade

web/src/modules/Student/ManageFees.tsx
Mudassar045: I thought to do this. Yeah that's better approach

web/src/modules/Student/ManageFees.tsx
```diff
@@ -218,6 +219,51 @@ class ManageFees extends Component<propTypes, S> {
...
+
+				if (fee_filter === "to_single_student" && student.section_id && student.id === selected_student_id) {
+					return true
+				}
+
+				// in case of empty selection
+				return false
+
+			})
+
```
taimur38: beautiful!

PR #238, opened 2020-03-19
Pagination

Mudassar045:
This PR contains
- [x] Convert student list to Type Script
- [x] Code refactoring
- [x] Pagination


web/src/modules/Student/List/index.tsx
```diff
@@ -0,0 +1,389 @@
...
+			items = items.filter(s => (s.tags === undefined || !s.tags["PROSPECTIVE"]))
+		}
+
+		if (forwardTo === 'payment') {
+			create = '/fees/manage'
+			createText = "Manage Fees"
+		}
+
+		const { students_per_page } = this.state
+		const card_items = items.slice(0, students_per_page)
```
taimur38, line 222: This is a good start, but here, instead of just making the page longer and longer, we should move on to the next "page".

There should only be PAGE_SIZE elements rendered at any time. Instead of having a variable called `students_per_page` I would have a variable called `page_index`. And this logic would be `const card_items = items.slice(page_index * PAGE_SIZE, (page_index + 1 * PAGE_SIZE))`

taimur38: Also would make sure that "search" still operates over all elements, and not just the current page

web/src/components/Printable/Student/list.tsx
```diff
@@ -11,18 +11,16 @@ type PropsTypes = {
...
 }
 
-type AugmentedStudent = MISStudent & { section: AugmentedSection }
-
 export const StudentPrintableList = (props: PropsTypes) => {
-    const students = props.studentClass !== "" ? 
-        props.students.sort((a, b) => (parseInt(a.RollNumber) || 0) - (parseInt(b.RollNumber) || 0)) : 
+    const students = props.studentClass !== "" ?
+        props.students.sort((a, b) => (parseInt(a.RollNumber) || 0) - (parseInt(b.RollNumber) || 0)) :
         props.students
```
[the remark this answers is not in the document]
Mudassar045: There are two cases:
- If there's no class selection, list will be sorted(all classes students) `Alphabetically`(by name) and by `classYear.
- If the class selected, students will be sorted alphabetically in parent component and by `rollNumber` in child component (print component)!

This should be streamlined and I will look into it.

web/src/modules/Student/List/index.tsx
Mudassar045: First, yes that's good way to use `page_index`
Second, Yes `search` and `filters` still operates over all elements.

PR #241, opened 2020-03-26
Profile card menu

Mudassar045:
This PR contains following:
- [ ] update `typescript: "3.8.3"` and `react-Scripts: "3.4.1"`
- [ ] add `react-bootstrap: "^1.0.0"`
- [x] dropdown-menu on profile cards with `delete student` and `print id card for single student`

**See Screenshot**:

2020-03-26-103951_389x194_scrot


web/src/components/Dropdown/style.css
```diff
@@ -16,7 +16,99 @@
...
+}
+
+.dropdown-menu:after {
+	top: -14px;
+	right: 10px;
+	left: auto;
+
+	border: 7px solid transparent;
+	border-bottom-color: #fff;
+}
```
taimur38, line 72: ok we used the react-bootstrap components - but why did we still have to do so much fancy css?

Mudassar045: react-bootstrap components requires bootstrap.css for styling. As we're not using `bootstrap.css` for styling, that's why we have to write own css.

taimur38: but then what did bootstrap give us?

Mudassar045: there are many things we can with `react-bootstrap`  but for now, We're using it for the `show/hide` of dropdown which handles gracefully (show/hide) inside side(within same component) out side clicks(in other components) .

I tried first to show/hide with local `state={ open: boolean }` for each component (student_card) which works but not the way we want. I was lacking the ability `to hide` on outside the component of dropdown.

In future, there are many things we will be doing with react-bootstraps, pretty sure it will help us.

taimur38: you can add an onclick listener to the body and an onclick listener to the modal in the modal component. when the correct one is clicked you just trigger onclose. that doesn't seem good enough to add the whole library

Mudassar045: I just be able to remove the `react-bootstrap`, `addEventListener()` help me out to handle to listen outside click. Now show/hide of modal working perfect.

PR #245, opened 2020-04-15
Corona module

taimur38:


web/src/actions/index.ts
```diff
@@ -919,4 +919,15 @@ export const resetFees = (students: MISStudent[]) => (dispatch: Function) => {
 	}, [])
 
 	dispatch(createMerges(merges))
+}
+
+export const saveStudentNeedyForm = (student: MISStudent) => (dispatch: Function) => {
```
taimur38, line 924: This is the exact same logic as the regular save student action no?

elixir/lib/sarkar/server/analytics.ex
```diff
@@ -1,240 +1,295 @@
...
-		IO.puts "wowwww"
-		req = :cowboy_req.reply(200, req)
-		{:ok, req, state}
-		
-	end
-
-	def init(%{bindings: %{type: "devices.csv"}} = req, state) do
-
-		{:ok, data} = case Sarkar.DB.Postgres.query(Sarkar.School.DB,
-		"SELECT school_id, client_id, to_timestamp(time/1000)::date::text as date, count(DISTINCT time)
```
taimur38, line 13: plz put this back to 'tabs' instead of spaces

web/src/constants/coronaQuetionnaire.ts
```diff
@@ -0,0 +1,77 @@
...
+		options: []
+	}
+]
+
+const UrduQuestions: FormQuestions[] = [
+	{
+		question: "",
+		options: ["", "", ""]
+	}
+]
```
taimur38, line 73: I assume these are not up to date?

taimur38: this is a smart way to organize the questions by language.

web/src/modules/Corona/index.tsx
```diff
@@ -0,0 +1,184 @@
...
+										.filter(s => this.state.filter.section_id && s.section_id === this.state.filter.section_id)
+										.sort((a, b) => parseInt(a.RollNumber) - parseInt(b.RollNumber))
+										.map(s => {
+
+											return <tr key={s.id}>
+												<td>{s.RollNumber}</td>
+												<td>{<Link to={`/student/${s.id}/profile`}>{s.Name}</Link>}</td>
+												<td>{s.ManName}</td>
+												<td>
+													<input type="checkbox" onChange={this.onCheckboxSelect(s)} />
```
taimur38, line 163: we should pull a value here as well. so that when a school owner sees this list, he can see who's information he has filled and who's he hasn't. so some sort of computed "value" function should be executed over each student. (ex: is neediness from 1-5 not undefined is good enough for a check)

web/src/modules/Corona/needy.tsx
```diff
@@ -0,0 +1,346 @@
...
+										<option value="GUARD">گارڈ</option>
+										<option value="FACTORY_WORKER">فیکٹری ورکر</option>
+										<option value="LABORER">مزدور</option>
+										<option value="MASON">مستری</option>
+										<option value="CARPENTER">کارپینٹر</option>
+										<option value="PAINTER">Painter</option>
+										<option value="MECHANIC">میکینک</option>
+										<option value="WELDER">ویلڈنگ</option>
+										<option value="OFFICE_BOY">کلرک</option>
+										<option value="DONT_KNOW">معلوم نہیں</option>
```
taimur38, line 293: there should be "Other" option with a text field to specify

web/src/modules/Corona/needy.tsx
```diff
@@ -0,0 +1,346 @@
...
+									</select>
+								</div>
+							}
+
+							{
+								this.state.needy_form.IncomeSource === "PRIVATE_JOB" &&
+								(this.state.needy_form.Occupation === "TEACHER" ||
```
taimur38, line 300: I would look into the former "check" logic, it could be easier to handle. for now this is fine, but if we add more complicated options in the future, you would want to use that

web/src/typings.d.ts
```diff
@@ -202,7 +202,7 @@ interface AugmentedSection {
 	faculty_id?: string
 }
 
-interface MISStudent {
+interface MISStudent extends NeedyForm {
```
taimur38, line 205: This is a shortcut, and is not really correct. What you could do, is make
`type MISStudent = { ... normal stuff } & NeedyForm` if you just want to merge them in together. But saying it extends needyform is not describing it correctly

web/src/constants/coronaQuetionnaire.ts
Mudassar045: Ohh I by mistake I added this file. Yes I was going to to this way but they were some input and questions dependencies for making form, so skip this approach.

web/src/actions/index.ts
Mudassar045: Yes that's the same action. I thought of different ways to do that, it was straightforward, I guess it should be like, and it's better to track
```elixir
export const saveStudentNeedyForm = (student_id: string, needy_form: NeedyForm) => (dispatch: Function) => {
	const merge = [
		{
			path: ["db", "students", student_id],
			value: needy_form
		}
	]
	
	dispatch(createMerges(merge))
}
```
@taimur38  do you say?

web/src/modules/Corona/index.tsx
Mudassar045: I thought the same way, but I would do that like
```tsx
<input type="checkbox" onChange={this.onCheckboxSelect(s)} checked={student.Needy || false} />
```
`student.Needy` gets only `true` when we save the form.

web/src/typings.d.ts
Mudassar045: yes correct. we don't need to extends it here like this.

web/src/modules/Corona/index.tsx
taimur38: ok

PR #247, opened 2020-04-22
handle reset/purchase code input using custom alert component like window.alert

Mudassar045:
This PR contains:
- [x] Replication of window [ `alert`, `prompt`, `confirm`]
- [x] Update logic to accept reset or purchase code using `MISPrompt`

**Activation Code for MIS**:
I think we should move our `package confirmation` logic to somewhere else, now it's only check for landing page, but mischool can be accessed via `URL`s other than `/landing` if trial is over. So we can move logic to:

- TrackedRoute Component (it's likely better option)
- Layout Component

What do you say @taimur38 ?


web/src/modules/Landing/index.js
```diff
@@ -206,10 +219,21 @@ class Landing extends Component {
 
 		return <Layout history={this.props.history}>
 			<div className="landing">
-				{!package_info.paid && package_info.date !== -1 && <div onClick={() => this.askForPassword()} className="trial-bar">
+				{!package_info.paid && package_info.date !== -1 && <div onClick={() => this.onActivationCodeModal()} className="trial-bar">
```
taimur38, line 222: On the trial, would they just be able to shut the activation code modal? or would it have the same behavior as today, where if they close it would reopen again?

Mudassar045: For `trial` they can open and close the activation code modal without any interruption.

Mudassar045:
> does this have the same behaviour as the previous modal when they close it?

No, it's a bit different than previous one. For the case of `trial-ended`,
- For previous modal
  `window.prompt` shows up in `componentDidMount`, so if the entered code is in correct or they cancel the prompt, we're doing
```js
if (accepted) {
	window.alert("ACCEPTED")
} else {
	window.location.reload()
}
```
- For current modal
It is based on `state` variable, so there's no `window.location.reload` in componentDidMount, just setting
```js
this.setState({ showPrompt: true })
```
So in this case, if they cancel or close the modal, it's never showed up till refresh or click on the alert bar (`Trial Period Ended`). That's bad solution

Both `window.prompt()` and `window.location.reload()` causing problem for syncing, To get rid both of them,  I purposed a solution to move our package confirmation logic somewhere else e.g. TrackedRoute. I'm also looking for other options.

taimur38:
I would just say that if the trial is over, it shouldn't be very easy to continue using the product without getting the trial extended or paying. I don't think they should just be able to close the modal and continue using

Mudassar045:
> I would just say that if the trial is over, it shouldn't be very easy to continue using the product without getting the trial extended or paying. I don't think they should just be able to close the modal and continue using

I handled through `TrackRoute` component, so whenever the trial ended, they will redirect to `verify-code` page. This help top avoid page/URL navigation to other component. Now it's working like a charm.

PR #256, opened 2020-07-16
make deduction from total amount

HumnaAleem:


web/src/modules/Expenses/Income-Expenditure/index.tsx
```diff
@@ -217,8 +217,7 @@ class IncomeExpenditure extends Component<propTypes, S> {
 							label: "EXPENSE",
 							date: expense.date
 						}
-
-					total_expense += amount
+					total_expense += (amount - deduction)
```
raoaliahmad, line 220: I don't see any difference in the salary expenses?

PR #260, opened 2020-07-30
students export to csv

Mudassar045:
Side work feature for MIS


web/src/modules/Exports/style.css
```diff
@@ -0,0 +1,55 @@
...
+	justify-content: center;
+}
+
+.exports-modal .row .button {
+	padding: 10px 15px;
+	margin-top: 10px;
+	margin-left: auto;
+}
+
+@media (max-width: 768px) {
```
taimur38, line 21: I like the designing for 3 screen widths instead of 2

web/src/modules/Student/List/index.tsx
```diff
@@ -13,6 +13,8 @@ import { StudenPrintableIDCardList } from 'components/Printable/Student/cardlist
 import getSectionsFromClasses from 'utils/getSectionsFromClasses'
 import { StudentIcon } from 'assets/icons'
 import { chunkify } from 'utils/chunkify'
+import Modal from 'components/Modal'
+import StudentExportModal from 'modules/Exports/studenExportModal'
```
taimur38, line 17: Should we put this in settings next to CSV import instead?

web/src/utils/downloadCSV.ts
```diff
@@ -11,4 +13,15 @@ const downloadCSV = (data: string[][], name: string) => {
 	hiddenElem.click();
 }
 
-export default downloadCSV;
\ No newline at end of file
+export const downloadAsCSV = (data: any, headers: string[], filename = "csv-generated-by-mischool") => {
```
taimur38, line 16: What's the difference between the 2 methods?

Mudassar045:
> Nice work - you made this on your own as a side task?

Yes

web/src/modules/Student/List/index.tsx
Mudassar045: Yes, exports should be on a same page.

web/src/utils/downloadCSV.ts
Mudassar045: Second method is using `papaparse` utility to generate csv using `data` and `headers` separately.

PR #261, opened 2020-08-06
admin staff password reset

Mudassar045:
This PR contains the functionality to send password to admin faculty after verification of phone number.


web/src/components/TrackedRoute/index.tsx
```diff
@@ -40,8 +41,18 @@ const TrackedRoute = ({ component, name, faculty_id, token, initialized, locatio
 		return <Redirect to="/verify-code" />
 	}
 
-	if (token && name) {
+	if (token && name && location.pathname === "/reset-password") {
```
taimur38, line 44: Isn't this going to mess up the old redirect logic - you've added an additional constraint here

web/src/modules/Password/index.tsx
```diff
@@ -0,0 +1,182 @@
...
+													{
+														adminFaculty
+															.sort((a, b) => a.Name.localeCompare(b.Name))
+															.map(f => <option key={f.id} value={f.id}>{f.Name}</option>)
+													}
+												</select>
+												<div className="">
+													<input type="text"
+														placeholder="Mobile number"
+														onBlur={(e) => setPhoneNumber(e.target.value)}
```
taimur38, line 130: Not onChange?

taimur38:
Ah never mind I see it's separate

web/src/modules/Password/index.tsx
Mudassar045: I use `onBlur` because the next action happen on clicking of `Send` button, so it's better way to mutate the state once on leaving the input field

web/src/components/TrackedRoute/index.tsx
Mudassar045: Tested and verified everything working fine. `location.pathname === "/reset-password"` is the guard for only reset-password component, so it will not mess up anything.

taimur38: Before it would redirect if just the name and token were there - now it's only redirecting if the name and token and pathname is reset-password -- the functionality changed here. I don't think you intended for that behavior

web/src/modules/Password/index.tsx
taimur38: I don't think I get it

Mudassar045: `onChange` event fired on every single character of input, so it mutates the state, to avoid this behaviour I use `onBlur` which only fire once the input field loses focus on click anywhere else and the state mutate once consequently.

But I guess there's problem with `react` working of `onChange`, here's what *MDN* says:
*For some elements, including <input type="text">, the change event doesn't fire until the control loses focus.*

I think, this statement, contradicts the working of `onChange` for react.

Mudassar045: Moreover here's the difference:
- The `onBlur` event is fired when you have moved away from an object without necessarily having changed its value.
- The `onChange` event is only called when you have changed the value of the field and it loses focus.

If react did work as described in above definition and on MDN, I would use `onChange`.

web/src/components/TrackedRoute/index.tsx
Mudassar045: So here's the breakdown of my intention:
```js
	if (token && name && location.pathname === "/reset-password") {
		return <Redirect to="/landing" />
	}

	if (token && location.pathname === "/reset-password") {
		return <Route {...rest} render={(props) => {
			trackRoute(location.pathname)
			return <Component {...props} />
		}} />
	}

	if (token && name) {
		if (faculty[faculty_id] === undefined) {
```
In these three `IFs`:
- First `if`, is used to prevent logged user to visit the URL `/reset-password`
- Second `if`, is used to navigate the `/reset-password` if token exists (`name` would be empty) and path is `/reset-password`
- Third `if` which handles all other cases, works only if above two IFs failed because these contain return statement

That's what I intended for.

PR #262, opened 2020-08-07
Feature/diary video links

HumnaAleem:


web/src/modules/Diary/index.tsx
```diff
@@ -301,6 +306,18 @@ class Diary extends Component<propTypes, S> {
...
+		const scrollY = window.pageYOffset
+		this.setState({showSubjects: true, scrollY: scrollY})
+		hideScroll()
+	}
+
+	handleToggleModal = () => {
+		this.setState({showSubjects: false})
+		showScroll()
+		window.scrollTo(0, this.state.scrollY)
+	}
```
taimur38, line 319: What is the goal with the scroll logic?

web/src/modules/Diary/SubjectModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+
+    const sorted_entries = getSortedEntries(lessons_data)
+
+    const copyLink = (link: string) => {
+        navigator.clipboard.writeText(link)
+    }
+
+    const getLink = (link: string) => {
+        const manipulatedLink = link.replace(/-/g, '/')
+        const finalLink = manipulatedLink.replace(/ /g, '%20').slice(0, -2)
```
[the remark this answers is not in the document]
HumnaAleem: replaced `replace funtion` wiith `encodeURI`

web/src/modules/Diary/SubjectModal.tsx
```diff
@@ -0,0 +1,215 @@
...
+    return { classTitles: [...class_titles], subjects: [...subjects] }
+}
+
+const getSortedEntries = (lessons_data: AugmentedIlmxLessons) => {
+    return Object.entries(lessons_data)
+        .sort(([, a], [_, b]) => b.watchCount - a.watchCount)
+}
+
+const getClassTitleFromLessonId = (lessonId: string): string => {
+    return lessonId.split("-")[1] || "nil"
```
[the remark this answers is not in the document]
HumnaAleem: No, It's just a string.

web/src/modules/Diary/index.tsx
HumnaAleem: If we don't use it. In mobile mode screen gets scroll below Modal.

web/src/modules/Diary/SubjectModal.tsx
```diff
@@ -0,0 +1,218 @@
...
+    lessonId: string
+    scrollY: number
+}
+
+const SubjectModal: React.FC<PropsType> = ({ events, lessons, fetchLessons, onClose }) => {
+
+    const [classFilter, setClassFilter] = useState('')
+    const [subjectFilter, setSubjectFilter] = useState('')
+
+    useEffect(() => {
```
raoaliahmad, line 27: overall looks good but you are fetching lessons every time they open up the modal so lets say if a teacher is opening the modal to select links for 4 different subjects it would request lessons 4 times which we don't need. So I would suggest moving the fetch lessons logic to the diary component so it only gets called once

HumnaAleem: done

PR #265, opened 2020-08-14
Feature/redesign fee module

HumnaAleem:


web/src/modules/Settings/ClassSettings/DefaultFee/index.tsx
```diff
@@ -1,45 +1,32 @@
...
 		this.state = {
-			fee: this.setDefaultFee(),
-			selected_class_id: "",
-			banner: {
-				active: false,
-				good: false,
-				text: ""
-			},
+			fee: props.settings.classes.defaultFee[props.classId] || this.setDefaultFee()
 		}
```
Mudassar045, line 24: My bad! I didn't realize that `props.settings.classses.defaultFee[props.classId]` would be empty when creating new class. So it should be:

```
const { settings, classId } = props
const fee = settings && settings.classes && settings.classes.defaultFee &&
settings.classes.defaultFee[classId]
this.state = {
	fee: fee || this.setDefaultFee()
}
```
This is to prevent to access undefined default fee against newly class id. Don't worry I added the fix in the `staging`. @HumnaAleem

PR #266, opened 2020-08-18
show attendance stats

HumnaAleem:


web/src/modules/Attendance/index.tsx
```diff
@@ -261,8 +261,33 @@ class Attendance extends Component<propTypes, S> {
...
 	}
 
+	getAttendanceStats = () => {
+		let attendanceStats = {
+			present: 0,
+			absent: 0,
+			leave: 0
+		}
+		Object.keys(this.state.selected_students)
+			.sort((id_a, id_b) => (this.props.students[id_a].RollNumber !== undefined && this.props.students[id_b].RollNumber !== undefined) && (parseFloat(this.props.students[id_a].RollNumber) - parseFloat(this.props.students[id_b].RollNumber)))
```
Mudassar045, line 271: You don't need to call `sort()` here! We don't need of it.

web/src/modules/Attendance/index.tsx
```diff
@@ -261,8 +261,33 @@ class Attendance extends Component<propTypes, S> {
...
+				const current_attendance = (x.attendance || {})[moment(this.state.date).format("YYYY-MM-DD")];
+				const status = current_attendance ? current_attendance.status : "n/a"
+				if (status === "PRESENT") {
+					attendanceStats.present++
+				} else if (status === "ABSENT") {
+					attendanceStats.absent++
+				} else if (status === "LEAVE" || status === "CASUAL_LEAVE" || status === "SHORT_LEAVE" || status === "SICK_LEAVE") {
+					attendanceStats.leave++
+				}
+			})
```
Mudassar045, line 283: There are two things you can do here:
- Use loop e.g. `for...of` instead of `map()` because you're not mapping anything here, just mutating the object `attedanceStats` of outer environment
- Use `reduce()` here, this is because you don't need to mutate the object of outer environment.

I will go with `reduce` which I think perfect solution.

web/src/modules/Attendance/index.tsx
```diff
@@ -324,9 +349,9 @@ class Attendance extends Component<propTypes, S> {
 				<div className="selectors">
 					<div className="row" style={{ flexWrap: "wrap" }}>
 						<div className="button select-all" onClick={this.selectAllOrNone}>{Object.values(this.state.selected_students).every(x => x) ? "Select None" : "Select All"}</div>
-						<div className="button select-all" onClick={this.selectPresentOrNone}>P</div>
-						<div className="button select-all" onClick={this.selectAbsentOrNone}>A</div>
-						<div className="button select-all" onClick={this.selectLeaveOrNone}>L</div>
+						<div className="button select-all" onClick={this.selectPresentOrNone}>P : {attendanceStats.present}</div>
+						<div className="button select-all" onClick={this.selectAbsentOrNone}>A : {attendanceStats.absent}</div>
+						<div className="button select-all" onClick={this.selectLeaveOrNone}>L :  {attendanceStats.leave}</div>
```
Mudassar045, line 354: Good thinking! Just show Ayesha, does it make sense to her.

web/src/modules/Attendance/index.tsx
HumnaAleem: I guess `for` will be better because we are just iterating not reducing an array.

web/src/modules/Attendance/index.tsx
HumnaAleem: Ok, removed

web/src/modules/Attendance/index.tsx
HumnaAleem: Yes, I asked her. She liked this idea too.

PR #272, opened 2020-09-01
Feature/set permissions

HumnaAleem:


web/src/modules/Landing/index.js
```diff
@@ -218,13 +218,12 @@ class Landing extends Component {
...
-		const teacher_expense_permission = permissions && permissions.expense ? permissions.expense.teacher : true;
-		const prospective_permission = permissions && permissions.prospective ? permissions.prospective.teacher : true;
-		const family_permission = permissions && permissions.family ? permissions.family.teacher : true;
-
+		const setupPage = user.permissions && user.permissions.setupPage;
+		const dailyStats = user.permissions && user.permissions.dailyStats;
+		const teacher_fee_permission = user.permissions && user.permissions.fee;
+		const teacher_expense_permission = user.permissions && user.permissions.expense;
+		const prospective_permission = user.permissions && user.permissions.prospective;
+		const family_permission = user.permissions && user.permissions.family;
```
Mudassar045, line 226: Do you think we should set `false` to all permission variables? Or `true` as we did earlier.

web/src/typings.d.ts
```diff
@@ -146,6 +146,8 @@ interface MISSettings {
 		dailyStats: { teacher: boolean }
 		setupPage: { teacher: boolean }
 		expense: { teacher: boolean }
+		family: { teacher: boolean }
+		prospective: {teacher: boolean}
 	}
```
Mudassar045: Do we need this anymore?

web/src/modules/Attendance/index.tsx
```diff
@@ -313,9 +312,9 @@ class Attendance extends Component<propTypes, S> {
 			messages,
 			return_link: window.location.href
 		});
-		const { settings, current_faculty, students, classes } = this.props;
+		const { current_faculty, students, classes } = this.props;
 		const isAdmin = current_faculty.Admin
-		const setupPage = settings.permissions && settings.permissions.setupPage ? settings.permissions.setupPage.teacher : true
+		const setupPage = current_faculty.permissions && current_faculty.permissions.setupPage
```
Mudassar045, line 317: What if current_faculty (logged in user) don't have `permissions` property ?

web/src/modules/Expenses/index.tsx
```diff
@@ -30,14 +29,13 @@ class ExpensePage extends Component <propTypes> {
 		// if student is new, just render the profile. when they hit save, redirect to the right page.
 		
 		const loc = this.props.location.pathname.split('/').slice(-1).pop();
-		const admin = this.props.user.Admin;
-		const permissions = this.props.permissions;
+		const {Admin: admin, permissions} = this.props.user;
```
Mudassar045, line 32: can you fix the code `alignment` ? This should be `{ Admin: admin, permissions }`

web/src/modules/Student/Single/index.js
```diff
@@ -52,5 +51,5 @@ class StudentPage extends Component {
 }
 export default connect(state => ({
 	user: state.db.faculty[state.auth.faculty_id],
-	permissions: state.db.settings.permissions,
+	// permissions: state.db.settings.permissions,
```
Mudassar045, line 54: If we don't getting permissions, just remove it instead of commenting out.

Mudassar045 (changes requested):
Everything seems good, but will take another look in the morning by pulling it on local machine and do some experiments.

web/src/modules/Expenses/index.tsx
HumnaAleem: done

web/src/modules/Student/Single/index.js
HumnaAleem: done

web/src/modules/Attendance/index.tsx
HumnaAleem: It will return undefined and show false case.

web/src/modules/Landing/index.js
HumnaAleem: I think we should set `false` to all permissions because the previous faculty shouldn't have the permission to perform admin tasks.

Mudassar045 (approved):

PR #275, opened 2020-09-07
Feature/add call btn

HumnaAleem:


web/src/modules/Teacher/Single/Create/index.tsx
```diff
@@ -443,7 +443,10 @@ class CreateTeacher extends Component<propTypes, S> {
 						<div className="divider">Contact Information</div>
 						<div className="row">
 							<label>Phone Number</label>
-							<input type="tel" {...this.former.super_handle(["Phone"], (num) => num.length <= 11)} placeholder="Phone Number" disabled={!canEdit} />
+							<div className="row" style={{ flexDirection: "row" }}>
+								<input style={{ width: "100%" }} type="tel" {...this.former.super_handle(["Phone"], (num) => num.length <= 11)} placeholder="Phone Number" disabled={!canEdit} />
+								{!this.isNew() && <a className="button blue call-link" href={`tel:${this.state.profile.Phone}`} > Call</a>}
+							</div>
```
Mudassar045, line 449: Can you make `call-button`  mobile only?

HumnaAleem: Reason? Aysha didn't mention that it should b only on mobile mode.

Mudassar045: I know, but does anyone can make call from website? I should be appropriate to hide in web view.

HumnaAleem: ok

PR #282, opened 2020-10-09
Feature/targeted instruction

HumnaAleem:


taimur38 (changes requested):
Lots of places where logic can be simplified.

The biggest area to improve is reducing the number of state variables we are keeping. If a state variable is a function of another state variable and / or props, then it does not need to be a state variable. This will simplify the logic all throughout the components.

We should also avoid using code that has side effects, and look to see where combinations of map / filter / reduce can more cleanly express the values that we are going for. Some examples are given

web/src/actions/index.ts
```diff
@@ -513,6 +513,25 @@ export const addPayment = (student: MISStudent, payment_id: string, amount: numb
...
+export const addReport = (studentId: string, report: Report, diagnostic_result: MISStudent['diagnostic_result']) => (dispatch: Function) => {
+	dispatch(createMerges([
+		{
+			path: ["db", "students", studentId, "report"],
+			value: report
+		},
+		{
+			path: ["db", "students", studentId, "diagnostic_result"],
+			value: diagnostic_result
+		}
```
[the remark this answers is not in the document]
taimur38: path should be [db, students, student_id, diagnostic_result, test_id] etc

web/src/typings.d.ts
```diff
@@ -55,8 +55,24 @@ interface RootDBState {
...
+	[learning_level: string]: LearningLevels
+}
+
+interface LearningLevels {
+	[lesson_id: string]: {
+		lesson_number: number
+		name: string 
+		description: string 
+		video_links: string[]
+		pdf_links: string
```
taimur38, line 75: I think we also need to have subject here

taimur38: and would need to add it to the csv template as well

taimur38 (commented):
looks good

web/src/modules/TargetedInstruction/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+					<select className="no-print" onChange={(e) => setType(e.target.value)}>
+						<option value="">Select Type</option>
+						<option value="Single Student">Single Student</option>
+						<option value="All Students">All Students</option>
+					</select>
+				</div>}
+				{(loc !== "test") &&
+					<>
+						{((type !== 'All Students' && loc === "report") || loc === "grades") && <div className="row">
+							<label className="no-print">Students</label>
```
taimur38, line 96: can simplify the logic I think.
```javascript
( (type !== 'All Students' && loc === 'report') || (loc === 'grades') ) && <div className="row">.... 
```

web/src/modules/TargetedInstruction/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+								{
+									Object.values(students)
+										.sort((a, b) => a.Name.localeCompare(b.Name))
+										.map((std) => (<option key={std.id} value={std.id}>{std.Name}</option>))
+								}
+							</select>
+						</div>}
+					</>
+				}
+				{loc === 'test' ? <Diagnostic label={pdfLabel} url={pdfUrl} /> :
```
taimur38, line 108: For a lot of the logic here, it may make more sense to set up `<Routes>` like we do in the other module. For example, look at student/single/index

web/src/modules/TargetedInstruction/index.tsx
```diff
@@ -0,0 +1,172 @@
...
+const getAllStudents = (sectionId: string, students: RootDBState["students"]) => {
+	return students = Object.values(students)
+		.reduce<RootDBState["students"]>((agg, student) => {
+			if (student.section_id === sectionId) {
+				return {
+					...agg,
+					[student.id]: student
+				}
+			}
+			return { ...agg, }
```
taimur38, line 154: can just return agg

web/src/modules/TargetedInstruction/style.css
```diff
@@ -0,0 +1,139 @@
...
+    display: flex;
+    flex-direction: row;
+    justify-content: space-between;
+    padding: 10px 0px;
+}
+.send-btn-div {
+    display: flex;
+    flex-direction: row;
+    justify-content: flex-end;
+}
```
taimur38, line 24: we must have existing classes which handle this...

web/src/utils/getSubjectsFromClasses.tsx
```diff
@@ -0,0 +1,18 @@
...
+                            ...agg2,
+                            sub
+                        ]
+                    }, [])
+            }
+        }, {})
+    return subjects;
+}
+
+export default getSubjectsFromClasses;
```
taimur38, line 18: Why do we need this? Can't you already just go through classes and get the subjects.

web/src/utils/targetedInstruction.tsx
```diff
@@ -0,0 +1,146 @@
+export const createReport = (students: RootDBState["students"], targeted_instruction: RootDBState["targeted_instruction"], testId: string) => {
```
taimur38, line 1: can you add comments which describe these functions - important for the util functions

web/src/utils/targetedInstruction.tsx
```diff
@@ -0,0 +1,146 @@
...
+    return Object.values(students)
+        .reduce((agg, std) => {
+            return {
+                ...agg,
+                [std.id]: {
+                    name: std.Name,
+                    report: Object.values((std.diagnostic_result && std.diagnostic_result[testId]) || {})
+                        .reduce((agg2: MISReport, { isCorrect, slo }) => {
+                            const category = targeted_instruction && targeted_instruction.slo_mapping[slo[0]] && targeted_instruction.slo_mapping[slo[0]].category;
```
taimur38, line 10: this only works if there is only one SLO for the question

taimur38:
```javascript
const categories = slo.map(s => targeted_instruction.slo_mapping[s].category)
```

web/src/utils/targetedInstruction.tsx
```diff
@@ -0,0 +1,146 @@
+export const createReport = (students: RootDBState["students"], targeted_instruction: RootDBState["targeted_instruction"], testId: string) => {
+    return Object.values(students)
+        .reduce((agg, std) => {
+            return {
+                ...agg,
+                [std.id]: {
+                    name: std.Name,
+                    report: Object.values((std.diagnostic_result && std.diagnostic_result[testId]) || {})
+                        .reduce((agg2: MISReport, { isCorrect, slo }) => {
```
taimur38, line 9: rename agg2 to report

web/src/utils/targetedInstruction.tsx
taimur38: I also think it would be better to just make a function called createReport which creates it for only ONE student. Then wherever you need to create a report, the component can do the work of doing the outer reduce for all students. you don't need to do it here

web/src/utils/targetedInstruction.tsx
taimur38:
```javascript
const augmented_slos = slo.map(s => { category: targeted_instruction.slo_mapping[s] })
const c = isCorrect ? 1 : 0 

return augmented_slos.reduce( (updated_report, curr_slo) => {
    const category = curr_slo.category;
    if(report[category]) {
        const prev = report[category]
        return {
            ...updated_report,
            [category]: {
                correct: prev.correct + c,
                possible: prev.possible + 1,
                link: curr_slo.link
            }
      }
   }

    return {
        ...updated_report,
        [category]: {
            correct: c,
            possible: 1,
            link: curr_slo.link
    }
}, report)
```

web/src/utils/targetedInstruction.tsx
```diff
@@ -0,0 +1,146 @@
...
+                                        link: targeted_instruction.slo_mapping[slo[0]] && targeted_instruction.slo_mapping[slo[0]].link
+                                    }
+                                }
+                            } else {
+                                return {
+                                    ...agg2,
+                                    [category]: {
+                                        correct: c,
+                                        possible: 1,
+                                        percentage: c / 1 * 100,
```
taimur38, line 28: percentage is a **derived** field. So you should not save it, rather whenever you want to use the percentage you can just calculate it from possible and correct.

web/src/utils/targetedInstruction.tsx
```diff
@@ -0,0 +1,146 @@
...
+        slo: '',
+        correct: obtained,
+        possible: total,
+        percentage: Math.round((obtained / total) * 100),
+        link: ''
+    }
+    return students
+}
+
+export const getAllStdData = (stdReport: Report) => {
```
taimur38, line 66: can you explain this function to me on phone

taimur38 (changes requested):
Just a few things to update. Looking much better though

web/src/constants/targetedInstruction.ts
```diff
@@ -0,0 +1,48 @@
...
+        sortable: true,
+
+    },
+    {
+        name: 'Percentage',
+        selector: 'percentage',
+        sortable: true,
+
+    }
+]
```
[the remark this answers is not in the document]
HumnaAleem: testing

web/src/modules/TargetedInstruction/index.tsx
HumnaAleem: If i do (type !== 'All Students' && loc === 'test')  . It doesn't work according to requirments

web/src/actions/core.ts
```diff
@@ -534,6 +534,7 @@ export const connected = () => (dispatch: (a: any) => any, getState: () => RootR
 				console.log("VERIFYY")
 				syncr.verify()
 
+				dispatch(fetchTargetedInstruction())
```
taimur38, line 537: this should be in src/index.js as:

```javascript
syncr.on('connect', () => dispatch(fetchTargetedInstruction())
```

web/src/actions/core.ts
```diff
@@ -641,4 +644,33 @@ export const fetchLessons = () => (dispatch: Dispatch, getState: () => RootReduc
 	})
 		.then(response => dispatch(getLessonsSuccess(response)))
 		.catch(err => dispatch(getLessonsFailure()))
+}
+
+export const getTargetedInstruction = () => ({
```
taimur38, line 649: all this logic should go in actions/index

web/src/actions/index.ts
```diff
@@ -1032,4 +1032,27 @@ export const sendTempPassword = (faculty: MISTeacher, password: string) => (disp
...
+			client_id: state.client_id
+		}
+	})
+		.then(response => dispatch({
+			type: "GET_TARGETED_INSTRUCTION_SUCCESS",
+			payload: response
+		}))
+		.catch(err => dispatch({
+			type: "GET_TARGETED_INSTRUCTION_FAILURE"
+		}))
```
taimur38, line 1057: make sure syncr is connected first, and if not, set this action to be dispached with `syncr.onNext('connect', () => dispatch(fetchTargetedInstruction()))`

web/src/modules/Landing/index.js
```diff
@@ -570,7 +570,7 @@ export default connect(state => ({
 	school_id: state.auth.school_id,
 	auth: state.auth,
 	client_id: state.client_id,
-	visible: state.db.targeted_instruction.visible,
+	visible: state.db.targeted_instruction_access,
```
taimur38, line 573: make sure you set this value if it is undefined.

HumnaAleem: yup, I set a false value by default.

web/src/actions/index.ts
HumnaAleem: ok

PR #286, opened 2020-10-20
Ilmx version

HumnaAleem:


web/src/modules/Front/index.js
```diff
@@ -479,7 +480,7 @@ class Front extends Component {
 				<div className="footer bg-red">
 					<div className="contact-us">
 						<div className="title">Contact Us</div>
-						<a href="tel:+923481112004">+92 348 111 2004</a>
+				<a href="tel:+923481112004">{helpLine.phoneAlt}</a>
```
Mudassar045, line 483: in `href` phone number not changed, it should be helpline.phoneInt (verify this)

web/src/modules/Student/Single/Create/index.tsx
```diff
@@ -632,86 +626,7 @@ class SingleStudent extends Component<propTypes, S> {
...
-							}
-							placeholder="Full Name"
-							disabled={!admin}
-						/>
-					</div>
-					<div className="row">
-						<label>Father Name</label>
-						<input type="text" {...this.former.super_handle(["ManName"])} placeholder="Father Name" disabled={!admin} />
-					</div>
-					<div className="row">
```
Mudassar045, line 652: when the user is `ILMX`, creating new student showing blank page.

web/src/modules/Teacher/Single/Create/index.tsx
```diff
@@ -337,7 +341,7 @@ class CreateTeacher extends Component<propTypes, S> {
 		return <div className="single-teacher-create">
 			{this.state.banner.active ? <Banner isGood={this.state.banner.good} text={this.state.banner.text} /> : false}
 
-			{this.isNew() && !this.state.toggleMoreInfo && <div className="form">
+			{this.props.ilmxUser && this.isNew() && !this.state.toggleMoreInfo && <div className="form">
 				<div className="divider">Personal Information</div>
```
Mudassar045, line 345: here the check should be `!this.props.ilmxUser`, if you don't do this, in normal user mode, showing all the fields.

web/src/modules/Teacher/Single/Create/index.tsx
```diff
@@ -358,17 +362,20 @@ class CreateTeacher extends Component<propTypes, S> {
...
-				this.isNew() && <div className="section-container" style={{ marginTop: "1.25rem" }}>
+				this.props.ilmxUser && this.isNew() && <div className="section-container" style={{ marginTop: "1.25rem" }}>
 					<div className="button green" onClick={this.toggleMoreInfo}>{this.state.toggleMoreInfo ? "Show Less Fields" : "Show Additional Fields"}</div>
 				</div>
 			}
 
 			{
-				(this.isNew() ? this.state.toggleMoreInfo : true) && <div className="form">
+				(this.props.ilmxUser && this.isNew() ? this.state.toggleMoreInfo : true) && <div className="form">
 					<div className="divider">Personal Information</div>
```
Mudassar045, line 379: Also it should be `!this.props.ilmxUser` for both places

Mudassar045 (changes requested):
A few changes required, but also we need to test it rigorously, so that both ilmxUser version and mis normal user version works fine.

web/src/modules/Analytics/index.tsx
```diff
@@ -23,22 +24,28 @@ class Analytics extends Component<P> {
...
-				<Route path="/analytics/exams" component={ExamsAnalytics} />
-				<Route path="/analytics/ilmexchange" component={IlmExchangeAnalytics} />
+				{
+					!this.props.ilmxUser && <> <Route path="/analytics/fees" component={Fees} />
+						<Route path="/analytics/attendance" component={Attendance} />
+						<Route path="/analytics/expenses" component={ExpenseAnalytics} />
+						<Route path="/analytics/teacher-attendance" component={TeacherAttendanceAnalytics} />
+						<Route path="/analytics/exams" component={ExamsAnalytics} />
+						<Route path="/analytics/ilmexchange" component={IlmExchangeAnalytics} />					</>
+				}
```
Mudassar045, line 45: We don't need to hide ilmxechange video analytics `/analytics/ilmexchange` for ilmx user. It will be accessible from landing page

web/src/modules/Student/Single/Create/index.tsx
HumnaAleem: resolved

web/src/modules/Teacher/Single/Create/index.tsx
HumnaAleem: There is no need to add ilmxUser check

web/src/modules/Teacher/Single/Create/index.tsx
HumnaAleem: There is no need to add ilmxUser check

web/src/modules/Analytics/index.tsx
HumnaAleem: done

Mudassar045 (approved):

PR #289, opened 2020-10-26
Feedback signup

HumnaAleem:


web/src/constants/generic.ts
```diff
@@ -4,10 +4,10 @@ const login_strategies = [
 	"Association",
 	"Edfin",
 	"School Referral",
-	"Individual",
+	"Area Manager",
 	"Ilm Exchange",
     "Facebook",
-    "other"
+    "Other"
 ]
```
Mudassar045, line 12: I think your `vs-code` linter is not working. We need to look into it.

web/src/modules/Front/SignUp.js
```diff
@@ -61,6 +63,7 @@ class SignUp extends Component {
 		if (this.state.otherLogin) {
 			const signup = { ...this.state.profile, typeOfLogin: this.state.otherLogin, date: moment.now() }
 			this.props.createSignup(signup)
+			return <Redirect to="/school-login" />
```
Mudassar045, line 66: I don't think it should be here. This should be in `render()`

web/src/modules/Front/SignUp.js
```diff
@@ -136,12 +140,21 @@ class SignUp extends Component {
...
+					<select {...this.former.super_handle(["profile", "city"])}>
+						<option value="">Select District</option>
+						{
+							Object.values(district_tehsils).map((district) => {
+								return district.map((dist)=> {
+									return <option key={dist} value={dist}>{dist && dist}</option>
+								})
+							})
+						}
+					</select>
```
Mudassar045, line 157: Did you confirm with Ayesha to show all the district for `PUNJAB` only? If she says show all districts, then you should display all instead of `PUNJAB`

web/src/modules/Front/SignUp.js
```diff
@@ -178,7 +191,7 @@ class SignUp extends Component {
 					</select>
 				</div>
 
-				<div className="button red" onClick={() => this.onSave()}> Submit</div>
+				<div className="button red" onClick={() => this.onSave()}> Create Login</div>
```
Mudassar045, line 194: I know Ayesha said this to be `Create Login` but it should `Signup` or `Signup for Free-Trial` because the form title is `Sign up`

web/src/utils/getDistrictTehsilList.ts
```diff
@@ -0,0 +1,829 @@
+export const getDistrictTehsilList = () => ({
+	"AZAD KASHMIR": {
+		BAGH: [
+			"BAGH",
+			"DHEERKOT"
+		],
+		BHIMBER: [
+			"BARNALA",
```
Mudassar045, line 8: This should be in constants directory, no?

Mudassar045 (changes requested):

web/src/modules/Front/SignUp.js
HumnaAleem: Yes, I asked her.

PR #290, opened 2020-11-02
Student mini form

HumnaAleem:


Mudassar045 (approved):
Everything is working fine. As we hide family ledger, please hide input field for family id for ilmx user. Also don't show certificates sections.

PR #291, opened 2020-11-03
Add referral info

HumnaAleem:


web/src/modules/Front/SignUp.js
```diff
@@ -148,9 +157,9 @@ class SignUp extends Component {
 					<select {...this.former.super_handle(["profile", "city"])}>
 						<option value="">Select District</option>
 						{
-								getDistricts().sort().map((dist)=> {
-									return <option key={dist} value={dist}>{toTitleCase(dist)}</option>
-								})
+							getDistricts().sort().map((dist)=> {
+								return <option key={dist} value={dist}>{toTitleCase(dist)}</option>
+							})
```
Mudassar045, line 162: Just mentioning we can write this like:

```jsx
getDistricts().sort().map(d => (<option key={d} value={d}>{toTitleCase(d)}</option>)
```

web/src/modules/Front/SignUp.js
```diff
@@ -129,11 +136,13 @@ class SignUp extends Component {
 	}
 
 	render() {
+
+	const Span = () => <span style={{ color: "red" }}>*</span>
+
```
Mudassar045, line 141: This is the small one (no performance issue), but function/components shouldn't be declared inside the render function. Some how, according to my knowledge. React keeps a version of function on each render. (https://stackoverflow.com/questions/41369296/react-functions-inside-render).

Mudassar045 (approved):
This is a quick one. I'm guessing, in next task (backend), make sure these state properties (naming) aligned with dashboard school login create properties names, so that we don't face any issue when we get the data on dashboard-side.

PR #292, opened 2020-11-05
send referral fields with signup

HumnaAleem:


web/src/actions/index.ts
```diff
@@ -238,13 +243,35 @@ export const createSignUp = (profile: Profile) => (dispatch: Function, getState:
...
+         	owner_other_job: "",
+         	association_name: "",     
+         	area_manager_name: "",
+         	computer_operator: "",
+         	owner_easypaisa_number: profile.ownerEasypaisaNumber,
+         	previous_software_name: "", 
+         	previous_management_system: ""
+		}
+	}
+
```
Mudassar045: I think your `eslint` is not working. Tomorrow we will fix it.

web/src/actions/index.ts
```diff
@@ -238,13 +243,35 @@ export const createSignUp = (profile: Profile) => (dispatch: Function, getState:
...
 	syncr.send({
 		type: "SIGN_UP",
 		client_type,
 		sign_up_id: v4(),
-		payload: {
-			...profile,
-		}
+		payload: signup_obj
```
Mudassar045, line 274: Second make sure on the backend, password don't get save to mis_signup table. Did you verify that?

web/src/modules/Front/SignUp.js
```diff
@@ -20,6 +20,7 @@ function emptyProfile() {
 		schoolName: "",
 		schoolPassword: "",
 		packageName: "Free-Trial",
+		
 		typeOfLogin: "",
 		referralSchoolName: "",
```
Mudassar045, line 25: As we migrated to this convention: `packageName: "FREE_TRIAL" | "TALEEM1" | "TALEEM2" | "TALEEM3"`
By default, `packageName` should be **FREE_TRIAL** instead of "Free-Trial". Also in the `<select>` option, put "FREE_TRIAL" as `<option>` value.

Mudassar045 (changes requested):
Everything looks fine, now in future we can easily expand the form fields. Just a few fixes needed to be changed.

Mudassar045 (approved):

PR #295, opened 2020-11-16
client device time check

Mudassar045:


web/src/modules/Landing/index.js
```diff
@@ -322,228 +322,231 @@ class Landing extends Component {
 				{!package_info.paid && package_info.date !== -1 && <div onClick={() => this.onActivationCodeModal()} className="trial-bar">
 					{this.getTrialWarningMessage()}
 				</div>}
-				<div className="horizontal-scroll-container">	
-			{this.props.ilmxUser === "ILMX" ?
+				{!device_time && <div onClick={() => this.onActivationCodeModal()} className="trial-bar">
+					Please set your device time and timezone correctly!
+				</div>}
```
taimur38, line 327: Why are we showing activation code modal on the timezone bar?

web/src/services/appService.ts
```diff
@@ -0,0 +1,45 @@
...
+
+const host = getOriginHttps()
+
+const baseURL = host + "/mis"
+
+const getServerTime = async (): Promise<any> => {
+	const opts = getRequestOptions()
+	const response = await fetch(`${baseURL}/server-time`, opts)
+	return handleResponse(response)
+}
```
taimur38, line 11: what's the benefit of splitting into so many functions.

```javascript
fetch(`{baseURL}/server-time`)
    .then(res => res.json())
```

taimur38: There is a lot of extra logic going on, and I am not sure what the benefits are of it.

web/src/services/appService.ts
```diff
@@ -0,0 +1,45 @@
...
+		let json_resp = {} as any
+
+		try {
+			json_resp = JSON.parse(text)
+		} catch (err) {
+			console.log(text)
+		}
+
+		if (!response.ok) {
+			const error = (json_resp && json_resp.message) || text || response.statusText || 'fetch-error'
```
taimur38, line 35: here is not where you check response.ok - response.text() would error... and you would use catch.

web/src/services/index.ts
```diff
@@ -0,0 +1 @@
+export * from './appService'
```
taimur38, line 1: Why create an index, and an appService file? what other kinds of services would we create?

I actually don't think this needs to be in a service file at all. We already have a utils folder for these kinds of things -- I think ideally you simply have utils/checkTimes.ts

web/src/utils/helpers.ts
```diff
@@ -4,14 +4,14 @@
  * @param time
 */
 
-export const getTimeString = (time: number): string => {
-	
+const getTimeString = (time: number): string => {
+
 	// using tilde(~) instead of Math.floor()
```
taimur38, line 9: why?

taimur38: it is much less readable

web/src/utils/helpers.ts
```diff
@@ -22,24 +22,40 @@ export const getTimeString = (time: number): string => {
 	return ret
 }
 
+/**
+ * description: utility to get the UTC time in milliseconds
+ * @param 
+ * @return big integer
```
taimur38, line 28: this descriptoin is not accurate... this function takes in a timestamp and returns a string showing hours minutes seconds..

web/src/utils/helpers.ts
```diff
@@ -22,24 +22,40 @@ export const getTimeString = (time: number): string => {
...
 	return localStorage.getItem("user")
+}
+
+
+export {
+	getTimeString,
+	getIlmxUser,
+	isMobile,
+	hideScroll,
+	showScroll
```
taimur38, line 60: Why have you started introducing this export style? What is the benefit

web/src/helpers/index.ts
```diff
@@ -0,0 +1 @@
+export * from './hostConfig'
```
taimur38, line 1: why have helpers / hostConfig? we already have utils

web/src/actions/index.ts
```diff
@@ -2,8 +2,12 @@ import { hash } from 'utils'
 import { createMerges, createDeletes, createLoginFail, analyticsEvent, uploadImages } from './core'
 import moment from 'moment'
 import { v4 } from "node-uuid"
-import Syncr from '@cerp/syncr';
-import { historicalPayment } from 'modules/Settings/HistoricalFees/historical-fee';
+import Syncr from '@cerp/syncr'
+import { Dispatch } from 'redux'
```
taimur38, line 6: typically just use Function to describe Dispatch instead of the type as the default does not handle thunk actions properly

taimur38 (changes requested):
I think there are many good ideas in here. But also a lot of over-engineering, and introducing lots of new things which don't need to be introduced. The logic for checking time is very simple - but through all the various files and function splitting, it is very tedious to follow.

web/src/actions/core.ts
```diff
@@ -2,6 +2,7 @@ import { Dispatch } from 'redux'
 import Syncr from '@cerp/syncr'
 import { loadDb } from 'utils/indexedDb';
 import { v4 } from 'node-uuid';
+import { getServerTime } from './index'
```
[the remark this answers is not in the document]
Mudassar045: Then I can `dispatch` from `web/src/idnex.js` where it makes more sense. like

```
syncr.on('connect', () => store.dispatch(getServerTime())
```

web/src/constants/actionTypes/index.ts
```diff
@@ -0,0 +1,9 @@
+enum AppActionTypes {
```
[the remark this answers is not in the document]
Mudassar045: No the name isn't being used by something else. This is to make distinction in future, among other redux actions types which we hard-coded or exported from  actions/index. For example, We have School, Teachers etc

```
enum SchoolActionTypes {
    SIGNUP = 'SCHOOL_SIGNUP', // this can be used for syncr action type
    SIGNUP_SUCCEED = 'SCHOOL_SIGNUP_SUCCEED'
    .
    .
    .
}
```
So there will be SoC (separation of concern) for what action is about for e.g. SchoolActionTypes, which makes more sense instead of `ActionTypes`

web/src/constants/actionTypes/index.ts
```diff
@@ -0,0 +1,9 @@
...
+	SERVER_TIME_REQUEST = "APP_SEVER_TIME_REQUEST",
+	SERVER_TIME_SUCCESS = "APP_SERVER_TIME_SUCCESS",
+	SERVER_TIME_FAILURE = "APP_SERVER_TIME_FAILURE"
+}
+
+export {
+	AppActionTypes
+}
```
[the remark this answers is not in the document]
Mudassar045: First, `export { }` like normal `named-export` e.g.  export enum AppActionTypes. As I explained below in one of comment, what I did like that. As we're using `TS` so it automatically detects exports from `index.ts` file so we can import from constants directly like:

```
// more concise way
import { AppActionTypes } from 'constants/index'

// lengthy way 
import { AppActionTypes } from 'constants/actionTypes/index' // no need to specify 'index' (TS knows)
```
Note: Ideally it should only be imported from `constants` only, but someone intentionally or unintentionally puts `constants` module as `node` dependency, so TS detect first from node_modules

web/src/modules/Landing/index.js
Mudassar045: That's my bad. onClick shouldn't be here.

web/src/services/appService.ts
Mudassar045: I was not sure too when I did this but the more practically use-case of this you can see in `branch-manager` where actually we're dealing with `https` endpoints. We're we have to write 'https://ilmexchange.com/mis/' for each `fetch-request`. We're pretty sure about the `host-origin` and prefix (in this case 'mis'). So don't you think, it makes more sense to split up instead of duplicating (I prefer duplications instead of unnecessary abstractions most of the time :p )

web/src/services/appService.ts
Mudassar045: I can take a look into it. I wanted to make it generic, I guess I missed it out.

web/src/services/index.ts
Mudassar045: I can think more about it. The reason behind doing this, we may be using some kind of services (different type of server endpoints, so it basically make no sense to put them in `utils` (we're not reusing anything if we do). I put like this, for only the future use-case.

web/src/utils/helpers.ts
Mudassar045: I wrote the reason in one of reply.

web/src/utils/helpers.ts
Mudassar045: This is unnecessary here. I forgot to remove it along with utility I wrote.

web/src/utils/helpers.ts
Mudassar045: Like there no perf benefits or some else. Instead of writing `export` on each single function definition, shouldn't be better to write `export` once and also if we follow this, we can take advantage of `renaming exports` which means we can introduce new export function name irrespective of what is previously written. For example

```
const foo = () => // returning something
const bar = () => // returning something

export {
      foo as helloWorld(),
      bar as doSomething()
} 
```
For that reason, I started practicing this.

web/src/helpers/index.ts
Mudassar045: Yeah you're right. I wasn't sure about this `helpers` directory. This all other places, I did stuff intentionally to carry out discussion about the patterns, conventions and some best practices for code splitting.

web/src/actions/index.ts
Mudassar045: Again that's my bad but "does not handle thunk actions properly", how a type `Dispatch`, can cause any issue to thunk?

Mudassar045:
> I think there are many good ideas in here. But also a lot of over-engineering, and introducing lots of new things which don't need to be introduced. The logic for checking time is very simple - but through all the various files and function splitting, it is very tedious to follow.

Thanks for looking into it. The sole intent purpose of writing down, splitting code and introduce new stuff, is to carry out discussion about some kind of conventions, coding best practices to follow while coding as team on a single or multiple projects. Code should belong to a team not to a single person. So anyone don't introduce new tedious ways (pardon me, I do :p), which reflects individual practice. This is ensure, the new joiners or anyone later can see clearly what to follow or what to not. That's all I want to say in above replies.

PR #296, opened 2020-11-30
Fix/minor changes

HumnaAleem:


web/src/modules/Diary/SubjectModal.tsx
```diff
@@ -101,72 +100,31 @@ const SubjectModal: React.FC<PropsType> = ({ events, lessons, onClose }) => {
...
     }
 
-    Object.entries(events || {})
-        .forEach(([, lessons_history]) => {
+    let lessonVideos
 
-            for (const [, item] of Object.entries(lessons_history)) {
+    for (let [id, lessonObj] of Object.entries(lessons_meta || {})) {
 
-                if (item.type === "VIDEO" && true) {
```
Mudassar045, line 127: As you see `if (item.type === "VIDEO" && true) {` you remove it. Lessons has other kind of data e.g. assessment, practice, it's better to filter out all lessons video of type only `VIDEO` in the parent component (diary module) and pass down to `subject modal`. This will make sure, on each time you open the modal, only type of video lessons there and have no performance issue. You can the profile flame graph, it's taking more than `6 seconds` which is insanely huge render time.
Screenshot from 2020-12-01 09-36-52

web/src/modules/Diary/SubjectModal.tsx
```diff
@@ -189,11 +147,6 @@ const getClassSubjectsInfo = (events: PropsType["events"]) => {
...
 }
 
-const getSortedEntries = (lessons_data: AugmentedIlmxLessons) => {
-    return Object.entries(lessons_data)
-        .sort(([, a], [_, b]) => b.watchCount - a.watchCount)
-}
-
 const getClassTitleFromLessonId = (lessonId: string): string => {
     return lessonId.split("-")[1] || "nil"
 }
```
Mudassar045, line 152: This is logic to get classes from lessonId.

web/src/modules/Diary/SubjectModal.tsx
```diff
@@ -101,72 +100,31 @@ const SubjectModal: React.FC<PropsType> = ({ events, lessons, onClose }) => {
...
+            lessonVideos = {
+                ...lessonVideos as IlmxLesson,
+                [id]: lessonObj
+            }
+        }
+    }
+    return lessonVideos
 }
 
 const getClassSubjectsInfo = (events: PropsType["events"]) => {
```
Mudassar045, line 130: You don't of this logic at all.

Mudassar045 (changes requested):

Mudassar045 (approved):

PR #307, opened 2021-02-04
Tip changes

HumnaAleem:


web/src/modules/TIP/Card.tsx
```diff
@@ -23,7 +25,9 @@ const Card: React.FC<P> = ({ class_name, teacher_name, school_name, subject }) =
                 <div className="flex flex-row justify-between w-full">
                     <div className="flex flex-col justify-center">
                         <div className="text-white text-lg">{teacher_name}</div>
-                        <div className="text-white text-base capitalize">{class_name.substring(0, 5) === "Level" ? `${group_name} Group` : class_name ? class_name : school_name}{subject && ` | ${subject}`}</div>                    </div>
+                        <div className="text-white text-base capitalize">{class_name.substring(0, 5) === "Level" ? `${group_name} Group` : class_name ? class_name : school_name}{subject && ` | ${subject}`}
+                            {(lesson_name && lesson_no) && ` | ${lesson_name.replace(/['"]+/g, '')} | ${lesson_no.replace(/['"]+/g, '')}`}</div>
```
taimur38, line 29: what does this do?

web/src/components/TrackedRoute/index.tsx
```diff
@@ -34,9 +34,9 @@ const TrackedRoute = ({
 }: propsType) => {
 	const Component = component
 
-	useEffect(() => {
-		window.scroll(0, 0)
-	})
+	// useEffect(() => {
+	// 	window.scroll(0, 0)
+	// })
```
taimur38, line 39: @Mudassar045 what were we trying to do again when we had this logic? We may need to adjust it - can show you one of the effects its having on our component

taimur38: I remember that we did it for a reason, but I don't remember for what

taimur38: instead of commenting this out, i think we just need to make sure the effect only fires when the route actually changes.

web/src/typings.d.ts
```diff
@@ -63,7 +63,7 @@ interface RootDBState {
 
 
 type TIPGrades = "1" | "2" | "3" | "KG" | "Oral Test"
-type TIPLearningGroups = "Blue" | "Yellow" | "Green" | "Orange" | "red"
+type TIPLearningGroups = "Blue" | "Yellow" | "Green" | "Orange" | "Black"
```
taimur38, line 66: we don't want a "black" group. We want the group to be called Oral

web/src/utils/TIP.tsx
```diff
@@ -252,7 +252,7 @@ export const convertLearningGradeToGroupName = (grade: TIPGrades) => {
 		"1": "Yellow",
 		"2": "Green",
 		"3": "Orange",
-		"Oral Test": "red"
+		"Oral Test": "Black"
```
taimur38, line 255: Again, no "Black" group. nobody will know what "black" group is... just call it oral

taimur38 (changes requested):

web/src/components/TrackedRoute/index.tsx
Mudassar045: This logic was for `scroll to top` after going to back to previous component (pages). Because in react, react-router preserve the scroll position. I have two solution for this:
**Solution A**
```
useEffect(() => {
		const unlisten = history.listen(() => {
			window.scrollTo(0, 0)
		})
		return () => {
			unlisten()
		}
	}, [history])
```
**Solution B**:
```
useEffect(() => {
	window.scrollTo(0, 0)	
}, [location.pathname])
```

web/src/typings.d.ts
HumnaAleem: If I change it to Oral It gets a white color => not visible for user

web/src/utils/TIP.tsx
HumnaAleem: Ok

web/src/modules/TIP/Card.tsx
HumnaAleem: In card, it gives info on different modules like.. in lesson plans it gives subject, lesson name ans lesson no .... in diagnostic test it gives class name, subject etc.

web/src/components/TrackedRoute/index.tsx
taimur38: I like solution b

web/src/typings.d.ts
taimur38: you've fixed this as per slack

PR #310, opened 2021-02-15
resolve minor issues

HumnaAleem:


taimur38 (commented):
one place where we can improve a nested ternary (x ? (a ? b : c) : d) into a simple function

PR #325, opened 2021-04-26
Added Expense pages and form to edit an expense

hamzaahmed00:
1. Added the Expense page
2. Added form to edit/add an expense


web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+					onChange={props.onChange}
+				/>
+			</div>
+		</div>
+	)
+}
+
+const index = (props: { match: { params: { id: any } } }) => {
+	let key: string = props.match.params.id
+	const expense: any = useSelector((state: RootReducerState) => state.db.expenses[key])
```
mudssrali, line 45: You can split this into two statements
```ts
   const { expenses } = useSelector((state: RootReducerState) => state.db)
   const expense = expenseId ? expenses[expenseId]: undefined
```

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+							title="Label"
+							defaultValue={state.label}
+							placeHolder="Please Write Description Here"
+							onChange={(e: { target: { value: any } }) =>
+								setState({
+									...state,
+									label: e.target.value
+								})
+							}
+						/>
```
mudssrali, line 140: We can simply use `<input ... />`, for now with already custom utilities

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+										className="rounded-full p-1 mt-1 text-gray-200 text-xs border-gray-200 border-2"
+										style={
+											state.category === cat
+												? { backgroundColor: '#1bb4bb' }
+												: { backgroundColor: 'transparent' }
+										}>
+										<h1 onClick={e => selectCategory(e)}>{cat}</h1>
+									</div>
+								)
+							})}
```
mudssrali, line 185: Always remember to have `key` when you're using `map()` as
```ts
categories.map(cat => (
     <div key={cat} ... />
))
```
Also as said above, avoid using `inline-styles`, you can use tailwindcss utilities. To generate className string conditionally, there's utility already setup that you can use as:

```ts
import clsx from 'clsx'
.
.
.
    <div className={clsx('rounded-full p-1 mt-1 text-gray-200 text-xs border-gray-200 border-2',
     state.category === cat ? 'bg-something' : 'bg-something')} />
```

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+								disabled={!isNew()}
+								defaultValue={moment(state.date).format('YYYY-MM-DD')}
+								className="w-full bg-transparent rounded border-2 border-blue-300 outline-none text-white placeholder-gray-400"
+								type="date"
+								onChange={(e: { target: { value: any } }) =>
+									setState({
+										...state,
+										date: new Date(e.target.value).getTime()
+									})
+								}
```
mudssrali, line 199:
```ts
onChange={(e) =>
	setState({
		...state,
		date: e.target.valueAsNumber // will give time in unix-timestamp
	})
}
```

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+							className="flex flex-1 flex-row justify-between mt-6 pl-4 pr-4 pt-2 pb-2 ml-1 mr-1 rounded-md"
+							style={{ backgroundColor: '#1BB4BB' }}>
+							<h1 className="text-xl text-gray-100 font-semibold">
+								{key === 'new' ? 'Add Expense' : 'Edit Expense'}
+							</h1>
+							<h1 className="text-xl text-gray-100 font-semibold">
+								Rs{' '}
+								{state.quantity !== undefined && state.amount !== undefined
+									? state.quantity * state.amount
+									: '0'}
```
mudssrali, line 219: Same goes here above `inline-styles` and `falsy` values

web/src/pages/expense/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+	const expense = useSelector((state: RootReducerState) => state.db.expenses)
+	const [groupedResults, setGroupedResults] = useState<any>(null)
+	const [totalExpense, setTotalExpense] = useState<number>(0)
+
+	const expenseFilter = (time: number, year: string, month: string): boolean => {
+		let objMonth = months[new Date(time).getMonth()]
+		let objYear = new Date(time).getFullYear()
+		if (objYear.toString() == year && objMonth == month) {
+			return true
+		}
```
mudssrali, line 29: also, `moment()` can be used here as well

web/src/pages/expense/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+		let total = 0
+		finalResult.forEach(element => {
+			Object.values(element).forEach(element => {
+				total = parseFloat(total.toString()) + parseFloat(element.amount.toString())
+			})
+		})
+
+		return total
+	}
+
```
mudssrali, line 83: Types can be extracted a single type and also here again use `reduce()` as said above

web/src/pages/expense/index.tsx
```diff
@@ -0,0 +1,158 @@
...
+					{groupedResults &&
+						groupedResults.map((data: any) => {
+							return (
+								<ExpenseCard
+									date={data[Object.keys(data)[0]].date}
+									expenseData={data}
+								/>
+							)
+						})}
+				</div>
```
mudssrali, line 144: It would be better to apply `groupedResults &&` on parent element. Also avoid using `any` and provide `key`

web/src/routing/index.tsx
```diff
@@ -188,6 +188,8 @@ export class Routes extends React.Component<RoutesProps, State> {
 						<Route exact path="/about-us" component={AboutUs} />
 						<Route exact path="/features" component={Feature} />
 						<Route exact path="/contact-us" component={ContactUs} />
+						<Route path="/expenses" exact component={Expense} />
+						<Route path="/expenses/:id" exact component={ExpenseForm} />
```
mudssrali, line 192: These are private routes, so you can use `<PrivateRoute />` instead of `<Route />`

mudssrali (changes requested):
Hi 👋 @hamzaahmed00 , Congratulations for your First PR 🥳 and looks good to me.
Good work in terms of layout designing and a few logic constructs. Appreciate your dedication and keeping flow in mind while working on expense module. Here is a few things that you work on them:
- Avoid using `inline-styles` and `any`
- Make your familiar with falsy values, `reduce()` and `moment()`
- Take care while naming a variable and and use of  typings
- Avoid using alert (it halts code execution), use toast instead

Also left a few questions in reviews comment, would expect your response on them.

web/src/routing/index.tsx
hamzaahmed00: Will do

web/src/pages/expense/index.tsx
hamzaahmed00: I will apply the groupedResults on the parent element but since we are creating a new type of data structure, I could not get it to work so I settled on any, I will look into this as well

web/src/components/cards/expense/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+const getYear = (date: number) => {
+	let d = new Date(date)
+	return d.getFullYear()
+}
+const getWeekDay = (date: number) => {
+	let d = new Date(date)
+	let days = ['SUN', 'MON', 'TUE', 'WED', 'THU', 'FRI', 'SAT']
+
+	return days[d.getDay()]
+}
```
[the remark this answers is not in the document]
hamzaahmed00: Isn't it time in milliseconds?

web/src/components/cards/expense/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+	totalDayExpense: number
+}
+
+const calculateExpense = (exp: any) => {
+	let total = 0
+	Object.values(exp).forEach((entry: any) => {
+		total = parseFloat(total.toString()) + parseFloat(String(entry.amount).toString())
+	})
+	return total
+}
```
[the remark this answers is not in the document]
hamzaahmed00: It was doing string concatenation so to be safe, I just parsed them into float by forcing them to convert to string

web/src/components/cards/expense/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+				<div className="flex flex-row items-center">
+					<h1 className="text-3xl font-bold mr-1">{getDay(date)}</h1>
+					<div>
+						<h1 style={{ fontSize: 10 }} className="font-bold text-gray-400">
+							{getMonth(date)}.{getYear(date)}
+						</h1>
+						<div className="p-1 bg-gray-400 rounded justify-center items-center text-center">
+							<h1 style={{ fontSize: 10 }} className="font-bold text-white">
+								{getWeekDay(date)}
+							</h1>
```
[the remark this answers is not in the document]
hamzaahmed00: I will do so

web/src/components/cards/expense/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+						</div>
+					</div>
+				</div>
+				<h1 style={{ color: '#40C0C6' }} className="font-medium">
+					Rs 0
+				</h1>
+				<h1 className="font-medium text-red-500">Rs {state.totalDayExpense}</h1>
+			</div>
+			{Object.entries(expenseData).map(([key, e]) => {
+				return (
```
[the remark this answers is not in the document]
hamzaahmed00: Got it

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+import { useDispatch, useSelector } from 'react-redux'
+
+interface addExpense {
+	label: string
+	amount: number
+	date: number
+	category: string
+	type: 'PAYMENT_GIVEN'
+	quantity: number
+}
```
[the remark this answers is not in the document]
hamzaahmed00: That expects an id but we are not generating the id here, that is why I just defined a type

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+			<div className="w-full rounded  focus:outline-none focus-within:outline-none mt-2">
+				<input
+					type="text"
+					disabled={props.disabled}
+					placeholder={props.placeHolder}
+					className="w-full bg-transparent rounded border-2 border-blue-300 outline-none text-white placeholder-gray-400"
+					defaultValue={props.defaultValue}
+					onChange={props.onChange}
+				/>
+			</div>
```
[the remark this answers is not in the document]
hamzaahmed00: Yes I have access to notion, I will look into it

web/src/components/cards/expense/index.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const calculateExpense = (exp: any) => {
+	let total = 0
+	Object.values(exp).forEach((entry: any) => {
+		total = parseFloat(total.toString()) + parseFloat(String(entry.amount).toString())
+	})
+	return total
+}
+
+const ExpenseCard: FC<propTypes> = ({ expenseData, date }) => {
```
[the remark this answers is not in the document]
hamzaahmed00: Okay got it

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+		}
+		if (
+			state.quantity <= 0 ||
+			state.amount <= 0 ||
+			state.quantity === undefined ||
+			state.amount === undefined
+		) {
+			alert('Quantity and Amount cannot be zero')
+			return
+		}
```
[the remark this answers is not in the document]
hamzaahmed00: Yes, they are integer and I will use toast from now on

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+		{
+			if (state.category === undefined) {
+				alert('Please select a Category')
+				return
+			}
+		}
+		if (isNew()) {
```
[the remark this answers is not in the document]
hamzaahmed00: Okay will do

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+		}
+	}
+
+	const categories = Object.keys(ExpenseCategories).filter(obj => obj.toString() !== 'SALARY')
+	const selectCategory = (e: any) => {
+		setState({
+			...state,
+			category: e.target.innerText
+		})
+	}
```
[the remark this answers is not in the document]
hamzaahmed00: I was getting the text that was written in the inner text and assigning it to the category

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+			setState({
+				amount: expense.amount,
+				category: expense.category,
+				date: expense.date,
+				label: expense.label,
+				quantity: expense.quantity,
+				type: 'PAYMENT_GIVEN'
+			})
+		}
+	}, [])
```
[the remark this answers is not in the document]
hamzaahmed00: I dont get this one

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+					defaultValue={props.defaultValue}
+					onChange={props.onChange}
+				/>
+			</div>
+		</div>
+	)
+}
+
+const index = (props: { match: { params: { id: any } } }) => {
+	let key: string = props.match.params.id
```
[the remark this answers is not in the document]
hamzaahmed00: That I forgot to change, I will do that

web/src/pages/expense/expenseForm/index.tsx
hamzaahmed00: Okay

web/src/pages/expense/expenseForm/index.tsx
hamzaahmed00: I just created this component so that I would not have to apply styling to all the other inputs over and over again

web/src/pages/expense/expenseForm/index.tsx
hamzaahmed00: Got it

web/src/pages/expense/index.tsx
hamzaahmed00: Got it

web/src/components/cards/expense/index.tsx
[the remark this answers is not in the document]
mudssrali: yes, it's! I just used for an example.

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+	return (
+		<div>
+			<h1 className="text-xl text-gray-100 font-normal ">{props.title}*</h1>
+			<div className="w-full rounded  focus:outline-none focus-within:outline-none mt-2">
+				<input
+					type="text"
+					disabled={props.disabled}
+					placeholder={props.placeHolder}
+					className="w-full bg-transparent rounded border-2 border-blue-300 outline-none text-white placeholder-gray-400"
+					defaultValue={props.defaultValue}
```
raoaliahmad, line 35: In React, `defaultValue` is used with uncontrolled form components whereas `value` is used with controlled form components
- `Controlled Components` form data is handled by a React component i.e using onChange
- `Uncontrolled Components` form data is handled by the DOM itself i.e can use a ref to get form values from the DOM

web/src/pages/expense/expenseForm/index.tsx
raoaliahmad: Instead of returning an object from the selector function i.e `state.db` and then getting expense out of it, only return expense i.e `state.db.expense` 
Because the way `useSelector` works, is that it compare the return value of the selector function with the return value from the prev render and then decides if it needs to re-render or not. It's official docs have really good explanation https://react-redux.js.org/api/hooks

web/src/pages/expense/expenseForm/index.tsx
```diff
@@ -0,0 +1,229 @@
...
+				/>
+			</div>
+		</div>
+	)
+}
+
+const index = (props: { match: { params: { id: any } } }) => {
+	let key: string = props.match.params.id
+	const expense: any = useSelector((state: RootReducerState) => state.db.expenses[key])
+	const [state, setState] = useState<addExpense>()
```
raoaliahmad, line 46: instead of leaving the state undefined what you can do is 
```
const initExpense = expense[id] ?? { defaultExpense Object here }
const [state, useState] = useState<addExpense>(initExpense)

useEffect(() => {
   if(expenseId && expense) {
       setState(expense[expenseId])
   }
},[expense])
```

raoaliahmad (changes requested):
Overall Good Work, 
Mudassir did a pretty good review so I didn't had much to comment on. 
PS: I have left some comments have a look at them.

web/src/pages/expense/expenseForm/index.tsx
hamzaahmed00: Why do we need useEffect now?

web/src/pages/expense/index.tsx
hamzaahmed00: moment for some reason refuses to work here and messes up my code so left this portion as is

web/src/pages/expense/salary/SalaryForm/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+import toast from 'react-hot-toast'
+import moment from 'moment'
+
+const SalaryForm = (props: {
+	location: { state: MISSalaryExpense[] }
+	match: { params: { id: any } }
+}) => {
+	const [salaries, setSalaries] = useState<MISSalaryExpense[]>(
+		props.location.state === undefined ? [] : props.location.state
+	)
```
mudssrali, line 21: why we're doing this?

web/src/pages/expense/salary/SalaryForm/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+	const [state, setState] = useState<Partial<MISSalaryExpense>>({
+		amount: parseFloat(teacher.Salary),
+		date: new Date().getTime()
+	})
+	const [detialsExpanded, setDetailsExpanded] = useState<boolean>(false)
+
+	const [type, setType] = useState<string>('Full')
+	const [reason, setReason] = useState<string>('')
+	const [totalDeductions, setTotalDeductions] = useState<number>(0)
+	const [totalPaid, setTotalPaid] = useState<number>(0)
```
mudssrali, line 35: here also we can create a single state variable using object

web/src/pages/expense/salary/SalaryForm/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+		})
+
+		setTotalDeductions(deduction)
+		setTotalPaid(paid)
+	}
+
+	useEffect(() => {
+		calculateDeductionsAndPaid()
+	}, [salaries])
+
```
mudssrali, line 56: First, why we need to calculate `deduction` and `paid` amount in useEffect()?
Second, we can use `reduce()` here and also and can eliminate multiple state setting by having a state object as. 

One thing to `note` here, actually we don't need to store these values in the state, we can just calculate them simply and use as result.
```tsx
	const { deduction, paid } = salaries.reduce(
		(agg, salary) => ({
			deduction: agg.deduction + salary.deduction,
			paid: agg.paid + salary.amount
		}),
		{ deduction: 0, paid: 0 }
	)
```
One thing more, if we have already something e.g. `salaries` as state variable, and we compute result from it, then we don't need to store that result in form of state variable. In short, don't derive state from state

web/src/pages/expense/salary/SalaryForm/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+					teacher.Name,
+					'PAYMENT_GIVEN',
+					teacher.id,
+					state.date,
+					0,
+					0,
+					'',
+					'SALARY'
+				)
+			)
```
mudssrali: I think we agreed on it to change to `addSalaryExpense()` to take an object as param instead of multiple arguments.

web/src/pages/expense/salary/SalaryForm/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+				setSalaries([
+					{
+						advance: 0,
+						amount: state.amount,
+						category: 'SALARY',
+						date: state.date,
+						expense: 'SALARY_EXPENSE',
+						type: 'PAYMENT_GIVEN',
+						deduction: 0,
+						deduction_reason: '',
```
mudssrali: we don't need of doing this, we can get the updated salaries from props, instead of saving them to state variable.

web/src/pages/expense/salary/SalaryForm/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+	}
+	const [offset, setOffset] = useState(0)
+
+	return (
+		<AppLayout>
+			<div className="flex flex-col-reverse lg:flex-row lg:m-14 lg:mx-32">
+				<div
+					ref={mainFormRef}
+					className="bg-gray-600  ml-3 mr-3 rounded-b-2xl flex flex-1 flex-col lg:mt-4 lg:rounded-2xl  lg:px-8">
+					<div className="m-5 text-gray-50 lg:space-y-5">
```
mudssrali, line 274: why we need of  `ref={mainFormRef}` here?

web/src/pages/expense/salary/SalaryForm/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+								onChange={text =>
+									setState({
+										...state,
+										amount: text.target.valueAsNumber
+									})
+								}
+							/>
+						</div>
+						{type !== 'Full' && (
+							<div>
```
mudssrali, line 350: You can create an `enum` of salary type as
```tsx
enum SalaryType {
FULL,
DEDUCTED,
ADVANCE
}
```
Now in your component
```tsx
const [salaryType, setSalaryType] = useState<SalaryType>()
.
.
.
{ type === SalaryType.FULL && (...) ...}
```

web/src/pages/expense/salary/SalaryForm/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+				enterTo="opacity-100 rotate-0 scale-100"
+				leave="transform duration-200 transition ease-in-out"
+				leaveFrom="opacity-100 rotate-0 scale-100 "
+				leaveTo="opacity-0 scale-95 ">
+				<div
+					style={{
+						top: detailsButtonRef?.current?.offsetTop || 0,
+						height: mainFormRef?.current?.offsetHeight + 100 || '50%',
+						width: mainFormRef?.current?.offsetWidth || '50%'
+					}}
```
mudssrali, line 473: What are we achieving here?

web/src/pages/expense/salary/SalaryForm/index.tsx
```diff
@@ -0,0 +1,410 @@
...
+					</div>
+					<div className="w-full flex flex-1 flex-col space-y-3  mt-2 overflow-y-scroll max-h-72">
+						<div className="flex flex-1 text-left ">
+							<div className="flex-1 font-medium">Month</div>
+							<div className="flex-1 font-medium text-center">Deductions</div>
+							<div className="flex-1 font-medium text-right">Paid</div>
+						</div>
+						{salaries ? (
+							salaries.map(entry => {
+								return (
```
mudssrali, line 497: `entry` can be changed to `salary` - naming convention

web/src/pages/expense/salary/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+	const [search, setSearch] = useState('')
+
+	useEffect(() => {
+		loadData()
+	}, [expenses, faculty])
+
+	const loadData = () => {
+		let groupsByFacultyID = Object.values(expenses)
+			.filter(element => element.expense === 'SALARY_EXPENSE')
+			.reduce((agg: { [id: string]: MISSalaryExpense[] }, curr: any) => {
```
mudssrali, line 25: You can do `filter()` part inside `reduce()` to some some computations.

web/src/pages/expense/salary/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+				return {
+					...agg,
+					[faculty_id]: [curr]
+				}
+			}, {} as { [id: string]: MISSalaryExpense[] })
+
+		console.log('Teachers', faculty)
+		console.log(groupsByFacultyID)
+		console.log('ID is ', groupsByFacultyID[Object.keys(groupsByFacultyID)[0]][0].faculty_id)
```
mudssrali, line 43: make sure, we remove these, anyhow will be removed in my commit

web/src/pages/expense/salary/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+					<div className="grid grid-cols-2 md:grid-cols-4 gap-4 md:gap-12 gap-y-12 md:gap-y-20">
+						{filteredStaff
+							.sort((a, b) => a.Name.localeCompare(b.Name))
+							.map(teacher => {
+								return (
+									<Link
+										to={{
+											pathname: `/salary/${teacher.id}`,
+											state: sortSalaries(teacherSalaries[teacher.id])
+										}}>
```
mudssrali, line 78: instead of creating route like `/salary/teacher.id`, we should do it as `staff/:id/salaries`

web/src/pages/expense/salary/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+						<div className="text-gray-500 text-xs md:text-base lg:text-lg">
+							{salaries ? salaries[0].amount : 0}
+						</div>
+					</div>
+					<div className="flex items-center justify-between flex-row">
+						<div className="text-gray-900 font-semibold">Deducted</div>
+						<div className="text-gray-500 text-xs md:text-base lg:text-lg">
+							{salaries ? salaries[0].deduction : 0}
+						</div>
+					</div>
```
mudssrali, line 123: Instead of showing `Last paid` or `Deducted` amount like this, you can pass them from parent component as prop variable which is more explicit what `salary[0].amount` means

web/src/pages/expense/salary/index.tsx
```diff
@@ -0,0 +1,154 @@
...
+					alt={teacher.Name || 'faculty'}
+				/>
+			</div>
+		</div>
+	)
+}
+
+export const isValidTeacher = (teacher: MISTeacher): boolean => {
+	return !!(teacher && teacher.id && teacher.Name)
+}
```
mudssrali, line 149: this already exist in `utils/index.ts`

web/src/routing/index.tsx
```diff
@@ -173,6 +175,10 @@ export class Routes extends React.Component<RoutesProps, State> {
 						<PrivateRoute path="/fees/add-historical-fee" component={HistoricalFee} />
 						<PrivateRoute path="/fee-menu" component={FeeMenu} />
 						<PrivateRoute path="/reset-password" component={ResetPassword} />
+						<PrivateRoute path="/expenses" exact component={Expense} />
+						<PrivateRoute path="/expenses/:id" exact component={ExpenseForm} />
+						<PrivateRoute path="/salary" exact component={Salary} />
+						<PrivateRoute path="/salary/:id" exact component={SalaryForm} />
```
mudssrali, line 181: Both these `/salary` and `salary/:id` should be changed to:

- `/staff/salaries` 
- `staff/:id/salaries`

web/tailwind.config.js
```diff
@@ -56,6 +56,9 @@ module.exports = {
 			},
 			screens: {
 				print: { raw: 'print' }
+			},
+			transitionProperty: {
+				height: 'max-height'
```
mudssrali, line 61: I didn't know about it before, but it works awesome <3

mudssrali (changes requested):
Good work, Keep up!
Here is a few areas that you need to consider while working and you can improve on them:
- Naming Convention
- Inferring types (implicit and explicit)
- Logic Constructs - JS or Pure ReactJS stuff
- New Language Features

I left a few questions also, please post your quick answer to them.

web/src/pages/expense/index.tsx
```diff
@@ -60,16 +114,50 @@ export const Expense = () => {
...
+			for (let k in a) {
+				key_a = k
+				break
+			}
+			for (let k in b) {
+				key_b = k
+				break
+			}
+			return a[key_a].date > b[key_b].date ? -1 : 1
+		})
```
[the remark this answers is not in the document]
hamzaahmed00: Because we are sorting the salaries (unsorted by default) from latest date to oldest date, this allows us to get the latest paid salary easily by doing salary[0] as it will always be the salary of the most recent date

web/src/pages/expense/salary/SalaryForm/index.tsx
hamzaahmed00: I was getting an undefined error for teachers who were never paid a salary without doing this

web/src/pages/expense/salary/SalaryForm/index.tsx
hamzaahmed00: I was doing it in useEffect so it could automatically recalculate whenever salaries changes

web/src/pages/expense/salary/SalaryForm/index.tsx
hamzaahmed00: We need that to calculate the exact height and width of this form and apply it to the pop up for past payments so that everything is properly aligned

web/src/pages/expense/salary/SalaryForm/index.tsx
hamzaahmed00: This is for the placement and sizing of the div that appears when we click "view past payments"

web/src/pages/expense/index.tsx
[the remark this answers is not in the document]
hamzaahmed00: Oh no, my mistake this is not for salaries, this actually sorts the expenses so that the latest day always shows up on top, it is not sorted by default so thats why I had to sort

web/src/pages/expense/index.tsx
```diff
@@ -18,10 +20,62 @@ export const Expense = () => {
...
+			)
+
+		setCategoryGroups(categoryGrouping)
+	}
+
+	const calculateCategoryExpense = (id: string) => {
+		return Object.values(categoryGroups[id]).reduce((agg: number, expense: any) => {
+			return agg + parseFloat(expense.amount.toString())
+		}, 0)
+	}
```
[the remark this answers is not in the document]
hamzaahmed00: if I don't specify type for expense, it does not let me get the amount from it

PR #326, opened 2021-04-27
Feature/quiz module

HumnaAleem:


web/src/modules/TIP/DetailedAnalysis/ClassView.tsx
```diff
@@ -35,7 +35,7 @@ const ClassView: React.FC<P> = ({ students, sorted_sections }) => {
 					))}
 				</div>
 			</div>
-			<div className="flex flex-col print:hidden">
+			<div className="flex flex-col print:hidden overflow-y-scroll h-60">
```
mudssrali: Instead of `overflow-y-scroll`, Wouldn't be good to use `overflow-y-auto` ?

web/src/modules/TIP/PDF.tsx
```diff
@@ -103,7 +103,7 @@ const PDF: React.FC<PropsType> = ({ match, targeted_instruction }) => {
 					<Link
 						className="no-underline"
 						to={
-							test_type === 'Diagnostic'
+							url[2] === 'diagnostic-test'
```
mudssrali: It would be good to have Enum of `TestTypes` instead of string-literals.

web/src/modules/TIP/DetailedAnalysis/ClassView.tsx
```diff
@@ -35,7 +35,7 @@ const ClassView: React.FC<P> = ({ students, sorted_sections }) => {
 					))}
 				</div>
 			</div>
-			<div className="flex flex-col print:hidden overflow-y-scroll h-60">
+			<div className="flex flex-col print:hidden overflow-y-scroll h-80">
```
mudssrali, line 38: Same goes here, to make `overflow-y-auto`, wouldn't be good to do so?

web/src/modules/TIP/LessonPlans/List.tsx
```diff
@@ -87,18 +87,18 @@ const List: React.FC<PropsType> = ({
...
-				let slo = Dynamic.get<string>(targeted_instruction, [
+				let quiz_id = Dynamic.get<string>(targeted_instruction, [
 					'curriculum',
 					class_name,
 					subject,
 					lesson_plan.lesson_number,
-					'slo'
-				])[0]
+					'quiz_id'
+				])
```
mudssrali, line 96: what is reason behind doing it so?

mudssrali (changes requested):
There's just a few quick questions and a suggestion for using `Enum`

web/src/modules/TIP/LessonPlans/List.tsx
HumnaAleem: we have changed the logic to get quizzes in lesson plans. we are mapping quiz against quiz id not slo

web/src/modules/TIP/DetailedAnalysis/ClassView.tsx
HumnaAleem: I have done but what's the difference ??

web/src/modules/TIP/PDF.tsx
HumnaAleem: I had to do this. because test type of diagnostic and oral test is Diagnostic.

mudssrali: I'm saying use `enum` of TestType as
```ts
enum TestType {
     DIAGNOSTIC = 'diagnostic-test'
     .
     .
     .
}
```

web/src/modules/TIP/DetailedAnalysis/ClassView.tsx
mudssrali: You should explore

HumnaAleem: got it, Thanks

web/src/modules/TIP/PDF.tsx
HumnaAleem: let test_type: TIPTestType = 'Diagnostic'
	if (url[2].indexOf('summative') >= 0) {
		test_type = 'Summative'
	}
	if (url[2].indexOf('formative') >= 0) {
		test_type = 'Formative'
	}
	if (url[2].indexOf('quizzes') >= 0) {
		test_type = 'Quiz'
	}
	I'm telling you I did this but I can't use test_type to routing. Because type of oral and diagnostic test is same

PR #340, opened 2021-05-24
Tab states maintained when navigating

hamzaahmed00:
Going back to the Home screen now results in the user to be in the tab from which he navigated instead of coming back to the Actions tab by default.

### DIFF
--- a/web/src/components/tabs/index.tsx
+++ b/web/src/components/tabs/index.tsx
@@ -1,16 +1,15 @@
 import React from 'react'
 import clsx from 'clsx'
 import { URLSearchParams } from 'url'
-import { useHistory } from 'react-router-dom'
 
 type TabbarProps = {
 	content: any
+	setTabParams?: (tab: number) => void
 	tab: any
 	setTab: (tab: any) => void
 }
 
-export const Tabbar = ({ tab, setTab, content }: TabbarProps) => {
-	const h = useHistory()
+export const Tabbar = ({ tab, setTab, content, setTabParams = () => {} }: TabbarProps) => {
 	return (
 		<div className="sticky inset-0 z-30 top-16 bg-teal-brand">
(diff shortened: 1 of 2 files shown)


web/src/components/tabs/index.tsx
```diff
@@ -8,13 +10,20 @@ type TabbarProps = {
...
 					<div
 						key={item.tab}
-						onClick={() => setTab(item.tab)}
+						onClick={() => {
+							setTab(item.tab)
+							h.push({
+								pathname: location.pathname,
+								search: '?activeTab=' + item.tab
+							})
+						}}
```
mudssrali, line 26: This component is being used in multiple places. Please pass an optional callback from parent component to push the `history`.

web/src/pages/home/index.tsx
```diff
@@ -31,7 +32,11 @@ const TabbarContent = [
 ]
 
 export const Home = () => {
-	const [activeTab, setActiveTab] = useState(Tabs.ACTIONS)
+	const urlParams = new URLSearchParams(location.search)
+
+	const [activeTab, setActiveTab] = useState(
+		parseInt(urlParams.get('activeTab') ?? '1') ?? Tabs.ACTIONS
+	)
```
mudssrali, line 39: Make sure, `activeTab` should be an `integer` instead of  `string`

mudssrali (changes requested):

web/src/pages/home/index.tsx
```diff
@@ -8,6 +8,8 @@ import { StatsTab } from './statistics'
 
 import { AppLayout } from 'components/Layout/appLayout'
 import { Tabbar } from 'components/tabs'
+import * as History from 'history'
```
raoaliahmad, line 11: What's this? This is not getting used anywhere.

raoaliahmad (approved):
Looks good, I saw an unused import I think should remove that.

PR #341, opened 2021-05-24
Restricted routing

hamzaahmed00:
Faculty can now navigate to only the tabs for which it has permission.

### DIFF
--- a/web/src/pages/home/actions.tsx
+++ b/web/src/pages/home/actions.tsx
@@ -10,6 +10,9 @@ import iconExams from 'assets/svgs/exams.svg'
 import iconAttendance from 'assets/svgs/attendance.svg'
 import iconDiary from 'assets/svgs/diary.svg'
 import iconExpense from 'assets/svgs/expense.svg'
+import { useSelector } from 'react-redux'
+import clsx from 'clsx'
+import toast from 'react-hot-toast'
 
 const links: CardProps[] = [
 	{
@@ -59,13 +62,30 @@ const links: CardProps[] = [
 	}
 ]
 
-export const ActionTab = () => {
+type PropTypes = {
+	permissions: {
+		fee: boolean
+		dailyStats: boolean
+		setupPage: boolean
+		expense: boolean
+		family: boolean
+		prospective: boolean
+	}
+	admin: boolean
+	subAdmin: boolean
+}
+
+export const ActionTab = ({ permissions, admin, subAdmin }: PropTypes) => {
 	return (
 		<div className="p-10 pt-6 mx-auto mb-10 md:w-full">
 			<div className="mb-6 text-lg text-center md:hidden">What would you like to do?</div>
 			<div className="grid grid-cols-2 gap-4 ">
 				{links.map((link, index) => (
-					<Card key={link.title + index} {...link} />
+					<Card
+						key={link.title + index}
+						{...link}
+						disabled={!checkPermission(permissions, link.title, subAdmin, admin)}
+					/>
 				))}
 			</div>
 			{isMobile() && (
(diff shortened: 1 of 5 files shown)


web/src/pages/home/actions.tsx
```diff
@@ -101,3 +129,29 @@ const Card = ({ title, icon, link }: CardProps) => {
 		</Link>
 	)
 }
+function checkPermission(
```
raoaliahmad, line 132: This looks like a function that can be re-used :clap: , should probably move this to utils and re-use this whenever you need to check permissions.

web/src/pages/home/index.tsx
```diff
@@ -34,11 +35,25 @@ export const Home = () => {
 	const [activeTab, setActiveTab] = useState(Tabs.ACTIONS)
 	const biggerThan880 = useMediaPredicate('(min-width: 880px)')
 
+	const { faculty } = useSelector((state: RootReducerState) => state.db)
```
raoaliahmad, line 38: MAKE SURE you only get the specific thing you need from the useSelector in this case `const faculty = useSelector((state: RootReducerState) => state.db.faculty)`

because useSelecctor compares the result with the result of the prev render and then decides if it needs to re-render or not. If you are taking out the db object. It will re-render whenever there's a change in the db. Which we don't need. we only want to re-render if the faculty changes. Should read through the following
https://react-redux.js.org/api/hooks

web/src/pages/home/actions.tsx
```diff
@@ -10,6 +10,9 @@ import iconExams from 'assets/svgs/exams.svg'
 import iconAttendance from 'assets/svgs/attendance.svg'
 import iconDiary from 'assets/svgs/diary.svg'
 import iconExpense from 'assets/svgs/expense.svg'
+import { useSelector } from 'react-redux'
```
raoaliahmad, line 13: is this being used anywhere?

web/src/pages/home/settings.tsx
```diff
@@ -18,37 +31,44 @@ export const SettingsTab = () => {
 		Object.values(students).filter(s => isValidStudent(s) && s.Active)
 	)
 
+	const setupPermission = admin || (subAdmin && permissions.setupPage)
```
raoaliahmad, line 34: here you can reuse that function `checkPermission` you made

web/src/pages/home/settings.tsx
```diff
@@ -9,7 +9,20 @@ import iconTeachers from './assets/teachers.svg'
...
+		expense: boolean
+		family: boolean
+		prospective: boolean
+	}
+	admin: boolean
+	subAdmin: boolean
+}
+
+export const SettingsTab = ({ permissions, admin, subAdmin }: PropTypes) => {
 	const { faculty, students, classes } = useSelector((state: RootReducerState) => state.db)
```
raoaliahmad, line 26: here also referencing the docs,
If you want to retrieve multiple values from the store, you can:
- Call useSelector() multiple times, with each call returning a single field value
- Use Reselect or a similar library to create a memoized selector that returns multiple values in one object, but only returns a new object when one of the values has changed.
- Use the shallowEqual function from React-Redux as the equalityFn argument to useSelector(),

web/src/pages/home/statistics.tsx
```diff
@@ -119,70 +133,87 @@ export const StatsTab = () => {
 		<div className="p-5">
 			<div className="mb-6 text-lg text-center md:hidden">Daily Statistics</div>
 
-			<div className="p-6 bg-white border border-gray-100 shadow-md rounded-2xl md:pt-3">
-				<div className="text-lg font-semibold text-center hidden md:block md:mb-4">
-					Daily Statistics
-				</div>
+			{admin || (subAdmin && permissions.dailyStats) ? (
```
raoaliahmad, line 136: reuse `checkPermission`

raoaliahmad (changes requested):
Overall good, some minor changes. 
- Make sure you replace permission checks with your reusable function. and that your function handles all the edge cases
- Update useSelectors to return specific values or pass the shallowequal function if you do want to retrieve multiple values.

PR #351, opened 2021-06-06
Testing feedback expense

hamzaahmed00:


web/src/components/cards/expense/index.tsx
```diff
@@ -16,6 +16,13 @@ type State = {
 
 const calculateExpense = (exp: { [id: string]: MISSalaryExpense | MISExpense }) => {
 	return Object.values(exp ?? {}).reduce((agg: number, entry: MISExpense | MISSalaryExpense) => {
+		if (entry.expense === 'MIS_EXPENSE') {
+			return (
+				agg +
+				parseFloat(String(entry.amount).toString()) *
+					parseFloat(String(entry.quantity).toString())
+			)
```
mudssrali, line 24: what is the purpose of `String(...).toString()` doing here?

web/src/components/cards/expense/index.tsx
```diff
@@ -57,22 +64,28 @@ const ExpenseCard = ({ date, expenseData, payments }: ExpenseCardProps) => {
...
-						</div>
+						{expense.expense === 'MIS_EXPENSE' ? (
+							<div className="flex flex-1 justify-end font-medium">
+								<h1>Rs {expense.amount * expense.quantity}</h1>
+							</div>
+						) : (
+							<div className="flex flex-1 justify-end font-medium">
+								<h1>Rs {expense.amount}</h1>
+							</div>
+						)}
```
mudssrali, line 88: it would be better to have an explicit parsing here.

web/src/pages/expense/form/index.tsx
```diff
@@ -63,6 +67,9 @@ export const ExpenseForm = ({ match }: RouteComponentProps<{ id: string }>) => {
 			})
 		)
 		toast.success('Expense entry has been updated')
+		setTimeout(() => {
+			history.goBack()
+		}, 1200)
```
mudssrali, line 72: In update case don't do this. May be user wants to update once more time.

web/src/pages/expense/form/index.tsx
```diff
@@ -125,25 +132,30 @@ export const ExpenseForm = ({ match }: RouteComponentProps<{ id: string }>) => {
...
+													? 'text-gray-200'
+													: 'text-gray-400 border-gray-400'
+											)}>
+											<h1
+												onClick={() =>
+													setState({ ...state, category: cat })
+												}>
+												{cat}
+											</h1>
+										</div>
```
mudssrali, line 156: You can make using a single `<button ... />` instead of wrapping <h1 ... /> inside a div.

web/src/pages/expense/index.tsx
```diff
@@ -82,7 +82,11 @@ export const Expense = () => {
 
 	const calculateCategoryExpense = (id: string) => {
 		return Object.values(state.categoryGroups[id]).reduce<number>((agg, expenses: any) => {
-			return agg + parseFloat(expenses.amount.toString())
+			return (
+				agg +
+				parseFloat(expenses.amount.toString()) *
+					parseFloat(expenses.quantity?.toString() ?? '1')
+			)
```
mudssrali, line 89: Now here we're not doing `String(...).toString()` why so?

web/src/pages/expense/index.tsx
```diff
@@ -177,7 +181,14 @@ export const Expense = () => {
...
 			Object.values(expenseItem).forEach(expenseItem => {
-				total = parseFloat(total.toString()) + parseFloat(expenseItem.amount.toString())
+				if (expenseItem.expense === 'MIS_EXPENSE') {
+					total =
+						parseFloat(total.toString()) +
+						parseFloat(expenseItem.amount.toString()) *
+							parseFloat(expenseItem.quantity.toString())
+				} else {
+					total = parseFloat(total.toString()) + parseFloat(expenseItem.amount.toString())
+				}
```
mudssrali, line 191: why you need to `parseFloat(total.toString())` do this ? total is already of type number

mudssrali (changes requested):

PR #387, opened 2021-07-08
Yusuf tweaks week of July 5th

yusufelley:
- Hide inactive users on staff login page
- Redesign activation page

### DIFF
--- a/web/src/pages/auth/ActivateMIS/index.tsx
+++ b/web/src/pages/auth/ActivateMIS/index.tsx
@@ -1,0 +1,205 @@
+import React, { useState } from 'react'
+import { connect } from 'react-redux'
+import { Redirect, Link } from 'react-router-dom'
+import { History } from 'history'
+import moment from 'moment'
+import { hash } from 'utils'
+import { resetTrial, markPurchased } from 'actions'
+import { AppLayout } from 'components/Layout/appLayout'
+import { ExclamationIcon, HappyEmojiIcon } from 'assets/icons'
+
+interface P {
+	initialized: boolean
+	schoolId: string
+	packageInfo: MISPackage
+	history: History
+
+	resetTrial: () => void
+	markAsPurchased: () => void
+}
+
+interface S {
+	code: string
+	isVerified: boolean
+	isPaid: boolean
+	isInvalidCode: boolean
+}
+
+const MISActivation: React.FC<P> = ({
+	initialized,
+	schoolId,
+	resetTrial,
+	markAsPurchased,
+	history
+}) => {
+	const [stateProps, setStateProps] = useState<S>({
+		code: '',
+		isInvalidCode: false,
+		isVerified: false,
+		isPaid: false
+	})
+
+	const verifyCode = async (code: string) => {
+		const reset_code = await hash(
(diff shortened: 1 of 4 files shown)


web/src/pages/auth/ActivateMIS/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+interface P {
+	initialized: boolean
+	schoolId: string
+	packageInfo: MISPackage
+	history: History
+
+	resetTrial: () => void
+	markAsPurchased: () => void
+}
```
mudssrali, line 19: We used this approach because we never integrated hooks before. Now we can get rid of props drilling via `connect()` and we don't need to wrap inside `connect()` because hooks can save us here.

Learn about react-redux hooks: [React-Redux Hooks](https://react-redux.js.org/api/hooks)

So what you can do here is:
```tsx
import { useSelector, useDispatch } from 'react-redux'

export const MISActivation = () => {
      const initialized = useSelector((state: RootReducerState) => state.initialized)
      const schoolId = useSelector((state: RootReducerState) => state.auth.school_id)
      const packageInfo = useSelector((state: RootReducerState) => state.db.package_info ?? { date: -1, trial_period: 15, paid: false })

.
.
.
}
```

web/src/pages/auth/ActivateMIS/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+		if (code === reset_code) {
+			resetTrial()
+			return [true, 'RESET']
+		}
+
+		if (code === purchase_code) {
+			markAsPurchased()
+			return [true, 'PURCHASED']
+		}
```
mudssrali, line 59: here you can use `useDispatch` hook instance to dispatch `resetTrial()` and `markAsPurchased` actions directly

```tsx
    const dispatch = useDispatch() // at the top inside component
   dispatch(resetTrial())
   .
   .
   .
   dispatch(markAsPurchased())
}
```

web/src/pages/auth/ActivateMIS/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+					...stateProps,
+					isInvalidCode: false
+				})
+			}
+		}, 3000)
+	}
+
+	if (!(schoolId && initialized)) {
+		return <Redirect to="/school-login" />
+	}
```
mudssrali, line 100: Good

web/src/pages/auth/ActivateMIS/index.tsx
```diff
@@ -0,0 +1,205 @@
...
+	(state: RootReducerState) => ({
+		initialized: state.initialized,
+		schoolId: state.auth.school_id,
+		packageInfo: state.db.package_info || { date: -1, trial_period: 15, paid: false }
+	}),
+	(dispatch: Function) => ({
+		resetTrial: () => dispatch(resetTrial()),
+		markAsPurchased: () => dispatch(markPurchased())
+	})
+)(MISActivation)
```
mudssrali, line 205: You can entirely get rid of this if you, use `react-redux hooks`

mudssrali (changes requested):

mudssrali (approved):
Thanks @yusufelley 👍🏻

PR #397, opened 2021-07-16
tweaks/yusuf/jul12

yusufelley:
- Make clickable link only on family card
- Add icon to distinguish between family and non-family students


web/src/assets/svgs/user-group.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" class="h-5 w-5" viewBox="0 0 20 20" fill="#fff" stroke="#6b7280">
+  <path d="M13 6a3 3 0 11-6 0 3 3 0 016 0zM18 8a2 2 0 11-4 0 2 2 0 014 0zM14 15a4 4 0 00-8 0v3h8v-3zM6 8a2 2 0 11-4 0 2 2 0 014 0zM16 18v-3a5.972 5.972 0 00-.75-2.906A3.005 3.005 0 0119 15v3h-3zM4.75 12.094A5.973 5.973 0 004 15v3H1v-3a3 3 0 013.75-2.906z" />
+</svg>
```
mudssrali, line 3: You can get rid of this.

web/src/pages/family/list.tsx
```diff
@@ -73,21 +67,28 @@ export const Family = ({ forwardTo, pageTitle }: FamilyProps) => {
...
 			<div className="bg-white rounded-xl lg:h-48  text-center border border-gray-50 shadow-md px-3 py-4 md:p-5">
-				<div className="font-bold pt-4 truncate w-4/5 mx-auto">
-					{toTitleCase(family.id)}
-				</div>
-				<div className="mt-2 space-y-0 text-sm md:text-base">
-					<CardItem title={'Gaurdian'} val={toTitleCase(family.name)} />
-					<CardItem title={'Phone'} val={toTitleCase(family.phone)} />
-					<CardItem title={'CNIC'} val={toTitleCase(family.ManCNIC)} />
-					<CardItem title={'Siblings'} val={Object.keys(family.students).length} />
-				</div>
```
mudssrali, line 90: We can wrap parent div with `<Link ... />` instead of parent div children

web/src/pages/students/list.tsx
```diff
@@ -21,6 +21,7 @@ import { checkStudentDuesReturning } from 'utils/checkStudentDues'
 import { addMultiplePayments } from 'actions'
 
 import UserIconSvg from 'assets/svgs/user.svg'
+import UserGroupSvg from 'assets/svgs/user-group.svg'
```
mudssrali, line 25: Instead of doing this you could import that svg like

```tsx
import {  UserGroupIcon } from '@heroicons/react/solid'
```
and can use like
```tsx
<UserGroupIcon className="w-4 h-4" />

**Review by mudssrali — CHANGES_REQUESTED** [review]

File: web/src/pages/staff/index.tsx [hunk 4]
```diff
@@ -7,7 +7,7 @@ import { AppLayout } from 'components/Layout/appLayout'
 import { CreateOrUpdateStaff } from './create'
 import { StaffMemberSalary } from 'pages/expense/salary/member'
 import { toTitleCase } from 'utils/toTitleCase'
-import Attendance from 'modules/Teacher/Single/Attendance'
+import { SingleStaffAttendance } from './attendance/single'
```
mudssrali, line 10: You committed unrelated changes as well. This will break

PR #400, opened 2021-07-23
redesign staff attendance

yusufelley:


web/src/utils/staffAttendanceMap.ts
```diff
@@ -0,0 +1,25 @@
...
+			case 'CASUAL_LEAVE':
+			case 'SICK_LEAVE':
+			case 'SHORT_LEAVE':
+			case 'PRIVILEGED_LEAVE':
+				res = 'Leave'
+				break
+			default:
+				break
+		}
+	})
```
mudssrali, line 23: We don't need to do this way. Also this is not a good use of `map()`. You can simply check object properties like we did while calculating attendance history. e.g.
```
let rest = ''
if((attendanceStatus.check_in) {
   res = 'Present'
}
.
.
.
```
Also you can get rid of `res` var by directly returning
```
let rest = ''
if((attendanceStatus.check_in) {
   return 'Present'
}
if(attendanceStatus.absent) {
  return 'Absent'
}
.
.
.
```

web/src/pages/staff/attendance/single/index.tsx
```diff
@@ -0,0 +1,481 @@
...
+	}
+
+	if (month !== '' && year === '') {
+		return moment(date).format('MMMM') === month
+	}
+
+	if (month !== '' && year !== '') {
+		return moment(date).format('MMMM') === month && moment(date).format('YYYY') === year
+	}
+}
```
mudssrali, line 481: We can extract to `utils/index.ts` because we're using it for students as well

mudssrali (changes requested):
It looks good to me. A few changes required and ultimately we can make  re-usable single attendance components for both staff and student.

web/src/components/Printable/Student/StudentAttendancePrint.tsx
```diff
@@ -70,3 +70,4 @@ export const StudentAttendancePrint = ({
 		</div>
 	)
 }
+	; ('')
```
mudssrali, line 73: is it intentionally or typo @yusufelley ?

mudssrali (approved):
Thanks @yusufelley

PR #404, opened 2021-07-29
Tweaks/yusuf/jul26

yusufelley:
- Add degree name field to staff profile
- Add custom categories to expenses


web/src/pages/expense/form/index.tsx
```diff
@@ -34,6 +34,26 @@ export const ExpenseForm = ({ match, history }: RouteComponentProps<{ id: string
...
+			state.category !== 'RENT' &&
+			state.category !== 'REPAIRS' &&
+			state.category !== 'STATIONERY' &&
+			state.category !== 'TAX' &&
+			state.category !== 'OTHER'
+		) {
+			setUsingCustomCat(true)
+			setViewCustomCat(true)
+		}
+	}, [state.category])
```
mudssrali, line 56: what actually this logic is doing?

web/src/pages/expense/form/index.tsx
```diff
@@ -171,7 +192,57 @@ export const ExpenseForm = ({ match, history }: RouteComponentProps<{ id: string
...
 											)
 										})}
+
+									{/* [+] Button */}
+									<button
+										onClick={() => {
+											!viewCustomCat && setState({ ...state, category: '' })
+											setViewCustomCat(prevState => !prevState)
+										}}
+										className={clsx(
```
mudssrali, line 202: we have already a component named as `<PlusButton ... />` (web/src/components/Button/plus.tsx), you could use that.

web/src/pages/staff/create.tsx
```diff
@@ -515,6 +516,18 @@ export const CreateOrUpdateStaff = () => {
 							</div>
 						</div>
 
+						<div className="flex flex-col space-y-1">
+							<div>Degree Name</div>
+							<textarea
+								name="DegreeName"
```
mudssrali, line 522: I think we should use `<input ... />` instead of `<textarea ... />`

mudssrali (changes requested):
All looks good @yusufelley Thanks. I was wondering that we should show custom categories as well when creating new expense item as:

```jsx
const getCategories = useCallback(() => {
   let Categories = new Set([])

  for (const e of Object.values(expenses)) {
	  Categories.add(e.category.trim().toUpperCase())
  }
  
  for (const [key] of Object.entries(ExpenseCategories)) {
	  Categories.add(key.trim())
  }
}, [expenses])
```
Note: we're importing `ExpenseCategories` from [web/src/constants/expense.ts](https://github.com/CERP/MIS/blob/de1a5f1f69392e3b12ffd06f880059b909f72ba2/web/src/constants/expense.tsx)

web/src/pages/expense/form/index.tsx
yusufelley: I used this so that the + button would not be highlighted at the same time as one of the categories. I also used it to display the custom category to user when viewing / editing an expense. I was trying to find a better way of going through the categories array and returning false if it matches, but I couldn't figure it out so I just left it as a bunch of statements.

PR #412, opened 2021-08-07
Zarish/tweaks and bug fixes

Zarish97:
### Added
- show pending amount on family card in payments
### Fixes
- fix password validation on signup (1 Upercase letter, 1 Number and at least 4 character long)
- close confirmation modal on removing sibling from family
- fix redirecting to families module on update or create new family


web/src/pages/family/single/index.tsx
```diff
@@ -160,11 +151,6 @@ export const SingleFamily = ({ match, location, history }: SingleFamilyProps) =>
 			)
 
 			toast.success('New family has been created')
-
-			// redirect to '/families'
-			return setTimeout(() => {
-				setState({ ...state, redirectTo: '/families' })
-			}, 1000)
 		}
```
mudssrali, line 154: We shouldn't remove this logic so the purpose of this logic is set `redirectTo` to familyes and also `return` if we don't do this, on creating new family both parts will be executed (if part and after if). We should keep this logic here.

web/src/pages/family/single/index.tsx
```diff
@@ -56,14 +55,6 @@ export const SingleFamily = ({ match, location, history }: SingleFamilyProps) =>
...
 	}, [classes])
 
-	// make sure if all students removed from family, redirect to families module
-	// to avoid redirect when famId = new, add isNewFam check
-	useEffect(() => {
-		if (!isNewFam && famId && siblings.length === 0) {
-			setState(prevState => ({ ...prevState, redirectTo: '/families' }))
-		}
-	}, [siblings, famId, isNewFam])
-
```
mudssrali, line 66: Yes, we don't need to do because we can derive behaviour for redirect/go-back from state variables directly inside component instead of using `useEffect()`

web/src/pages/family/single/index.tsx
```diff
@@ -233,7 +219,7 @@ export const SingleFamily = ({ match, location, history }: SingleFamilyProps) =>
 		toast.success(student.Name + ' has been removed')
 	}
 
-	if (state.redirectTo) {
+	if (!isNewFam && famId && siblings.length === 0) {
```
mudssrali, line 225: Here we can  add `|| redirectTo`, to goback() on create new family. so
```jsx
if((isNewFam && state.redirectTo) || (!isNewFam && famId && siblings.length === 0)) {
   history.goBack()
}
```

web/src/utils/helpers.ts
```diff
@@ -46,7 +46,7 @@ export const getIlmxUser = (): string => {
 
 export const isValidPassword = (password: string) => {
 	// eslint-disable-next-line
-	const regex = /(?=.*[a-z0-9])(?=.*[A-Z0-9])(?=.{4,})/g
+	const regex = /^(?=.*[a-z])(?=.*[A-Z])(?=.*\d)[a-zA-Z\d]{4,}$/gm
```
mudssrali, line 49: Good catch!

web/src/pages/family/list.tsx
```diff
@@ -71,6 +74,21 @@ type CardProps = {
 }
 
 const Card = ({ family, forwardTo }: CardProps) => {
+	const { students } = family
+	const siblings = Object.values(students)
```
mudssrali, line 78: we can directly do
```jsx
const siblings = Object.values(family.students ?? {})
```
It is good practice to always has a `guard-condition` which prevent unexpected run-time bugs while working with Object methods e.g. Object.values(), Object.entries() or Object.keys()

web/src/pages/family/list.tsx
```diff
@@ -114,10 +136,17 @@ const Card = ({ family, forwardTo }: CardProps) => {
...
+	<div className="flex items-center justify-between flex-row text-sm ">
 		<div className="font-semibold">{title}</div>
-		<div className="text-gray-500 text-xs">{val}</div>
+		<div
+			className={clsx(
+				{ 'text-red-500': pending, 'text-gray-500': !pending },
+				`text-right text-xs`
+			)}>
+			{val}
+		</div>
```
mudssrali, line 150: Here we could do simply
```
clsx('text-right text-xs', pending ? 'text-red-500' : 'text-gray-500')
```

mudssrali (changes requested):
Congratulations on your first PR @Zarish97 into mischool-codebase. A few comments that you can take a look into them.

web/src/pages/family/list.tsx
```diff
@@ -74,18 +72,17 @@ type CardProps = {
 }
 
 const Card = ({ family, forwardTo }: CardProps) => {
-	const { students } = family
-	const siblings = Object.values(students)
+	const siblings = Object.values(family.students ?? {})
 	const siblingPayments = useCallback(() => {
 		const payments = siblings.reduce(
-			(agg, sibling) => [...agg, ...Object.values(sibling.payments)],
+			(agg, sibling) => [...agg, ...Object.values(sibling.payments ?? {})],
```
mudssrali, line 78: Good to see this here. We have a few schools for which students have `undefined` payments.

web/src/pages/family/single/index.tsx
```diff
@@ -26,13 +26,14 @@ type State = Partial<MISStudent> & {
...
 	}
+	redirectTo?: String
 }
 
 export const SingleFamily = ({ match, location, history }: SingleFamilyProps) => {
 	const dispatch = useDispatch()
 	const { students, classes } = useSelector((state: RootReducerState) => state.db)
 	const { ref, setIsComponentVisible, isComponentVisible } = useComponentVisible(false)
-	const [studnetToDelete, setStudentToDelete] = useState<MISStudent>()
+	const [studentToDelete, setStudentToDelete] = useState<MISStudent>()
```
mudssrali, line 36: Thanks for fixing `typo` :p

mudssrali (approved):

PR #414, opened 2021-08-09
Desktop designs for Student and Family payments

hmzahmed:
## Added
- Add desktop redesign of student payments page
- Add desktop  redesign of family Payments page


mudssrali (changes requested):
Thanks @hmzahmed , everything looks for both `single-student` and `single-family` payments. A few things should be fixed:
- both student and family payment pages should have margin bottom
- Siblings card for family, should have small card item (name and picture)  instead of big picture and only name.
- Make all filters to fit-width equally on cards
- Make sure each component should have `unique-key prop` and get rid of consoles

PR #448, opened 2021-09-26
add custom date filter

ZarishIqbal:
The analytics can be filtered for a custom date range.


web/src/utils/analytics.ts
```diff
@@ -158,3 +160,13 @@ export const getDefaultExpenseStats = () => ({ INCOME: 0, EXPENSE: 0, PROFIT: 0
 export const getExpenseType = (payment: MISStudentPayment | BaseMISExpense) => {
 	return payment.type === 'SUBMITTED' ? ExpenseGraphTypes.INCOME : ExpenseGraphTypes.EXPENSE
 }
+
+export const toGraphTitle = (title: string) => {
+	if (title === DateTypes.LAST_7_DAYS) {
+		return `${toTitleCase(title)}' Graph`
+	} else if (title === DateTypes.CUSTOM) {
+		return `${toTitleCase(title)} Graph`
+	}
```
mudssrali, line 169: we can get rid of `else if`, simply do
```tsx
if(titile === DateTypes.CUSTOM)
...
```

PR #451, opened 2021-09-29
add password modal to starting test

ZarishIqbal:
Added password protection modal to starting tests.
Design changes


web/src/components/TIP/test-password-modal.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+export const TestPasswordModal = ({ testType, setRedirect }: TestPasswordModalProps) => {
+	const [password, setPassword] = useState('')
+	const [error, setError] = useState(false)
+
+	const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
+		event.preventDefault()
+		if (testPasswordMapping[testType] === password) {
+			return setRedirect(true)
+		}
```
mudssrali, line 27: here we also need to dispatch an action to set test password and based on test password, we have show test after redirection. 
There are two possible for it:
-  Embed password for test while redirecting
-  Store password in redux store -> temporary

mudssrali (changes requested):
Overall it looks good - Thanks @ZarishIqbal . A few things in the comments that we should follow to enhance code-readability, to follow best practices.

mudssrali (approved):
Thanks @ZarishIqbal - Requested changes looks fine to me.

PR #453, opened 2021-10-02
add multiple passwords to Starting, Oral and Summative tests

ZarishIqbal:
Added multiple password protection to diagnostic, oral and summative tests.


web/src/types/typings.d.ts
```diff
@@ -355,7 +355,10 @@ interface RootReducerState {
 		tests: TIPTests
 		slo_mapping: SLOMapping
 		curriculum: TIPCurriculum
-		test_password: string
+		test_password: {
+			[type: string]: string
+			[password: string]: string
+		}
```
mudssrali, line 361: type typings would be `TIPTestType` and we can simplify like
```tsx
test_password: {
    type: TIPTestType
    password: string
}
```

web/src/utils/helpers.ts
```diff
@@ -76,3 +76,6 @@ export const getNumberArrayFromRange = (length: number, start = 0) => {
 	}
 	return Array.from(Array(length), (x, i) => i + start)
 }
+export const checkPasswordTestType = (testType: string) => {
```
mudssrali, line 79: again testType would be of type `TIPTestType`
```tsx
export const checkPasswordTestType = (type: TIPTestType) => {
...
}
```

web/src/utils/indexedDb.ts
```diff
@@ -79,7 +79,10 @@ export const initState: RootReducerState = {
 		tests: {},
 		slo_mapping: {},
 		curriculum: {} as TIPCurriculum,
-		test_password: ''
+		test_password: {
+			type: '',
+			password: ''
+		}
```
mudssrali, line 85: If we're going to change type of `type` from `string` to `TIPTestType`, we might need do trick here.

web/src/utils/localStorage.ts
```diff
@@ -78,7 +78,10 @@ const initState: RootReducerState = {
 		tests: {},
 		slo_mapping: {},
 		curriculum: {} as TIPCurriculum,
-		test_password: ''
+		test_password: {
+			type: '',
+			password: ''
+		}
```
mudssrali, line 84: same goes here.

web/src/components/TIP/test-password-modal.tsx
```diff
@@ -24,13 +24,15 @@ export const TestPasswordModal = ({ testType, setRedirect }: TestPasswordModalPr
 	const dispatch = useDispatch()
 	const handleSubmit = (event: React.FormEvent<HTMLFormElement>) => {
 		event.preventDefault()
-		if (testPasswordMapping[testType] === state.password) {
+
+		if (testPasswordMapping[testType][state.password]) {
 			dispatch({
 				type: ActionTypes.TIP_TEST_PASSWORD,
-				data: state.password
+				payload: { type: testType, password: state.password }
```
mudssrali, line 31: To just share, `data` and `payload` have both same meaning in term of usage. You can literally use any other variable than them but be careful when getting value inside a reducer.

web/src/components/TIP/test-password-modal.tsx
```diff
@@ -8,14 +8,14 @@ type TestPasswordModalProps = {
...
 const testPasswordMapping = {
-	Diagnostic: 'abc1',
-	Formative: 'abc2',
-	Summative: 'abc3',
-	Oral: 'abc4'
+	Diagnostic: { abc1: true, abc2: true },
+	Formative: { abc3: true, abc4: true },
+	Summative: { abc3: true, abc4: true },
+	Oral: { abc5: true, abc6: true },
+	Quiz: { abc3: true, abc4: true }
```
mudssrali, line 18: Here we might need to change passwords for `Formative` and `Quiz`, no?

mudssrali (changes requested):
Thanks @ZarishIqbal Looks good to me. A few comments about test `type` typings and password.

web/src/components/TIP/test-password-modal.tsx
ZarishIqbal: Since quizzes and formative tests aren't password-protected yet, their password doesn't matter for now. I'll remove these.

web/src/components/TIP/test-password-modal.tsx
```diff
@@ -12,11 +12,9 @@ type PasswordMapping = {
 }
 const testPasswordMapping = {
 	Diagnostic: { abc1: true, abc2: true },
-	Formative: { abc3: true, abc4: true },
 	Summative: { abc3: true, abc4: true },
-	Oral: { abc5: true, abc6: true },
-	Quiz: { abc3: true, abc4: true }
-} as PasswordMapping
+	Oral: { abc5: true, abc6: true }
+} as Partial<PasswordMapping>
```
mudssrali, line 17: makes sense, nice

mudssrali (approved):
Thanks @ZarishIqbal!

PR #472, opened 2021-11-16
add lesson plan print all flow

ZarishIqbal:
download all lesson plans


web/src/modules/TIP/LessonPlans/List.tsx
```diff
@@ -160,6 +160,11 @@ const List = ({ match, history }: RouteComponentProps) => {
 
 	const lesson_progress = getLessonProgress(faculty[facultyId])
 
+	const printAll = () => {
+		Object.entries(lessonPlanBooklets[class_name] ?? {}).forEach(([subject, link]) => {
+			downloadLink(`${class_name}_${subject}.pdf`, link, printRef)
+		})
```
mudssrali, line 166: There are twelve combinations we have for each group. Lesson pdf should be printed for visited class/subject, not for all class' subjects, no?

web/src/modules/TIP/TIPConstants.ts
```diff
@@ -5,3 +5,40 @@ export enum Types {
 	SINGLE_QUIZ_VIEW,
 	SINGLE_SLO_VIEW
 }
+type booklets = {
+	[level in TIPLevels]: { [subject in TIPSubjects]: string }
+}
+export const lessonPlanBooklets: Partial<booklets> = {
+	'Level KG': {
+		English:
+			'https://storage.googleapis.com/targeted-instructions/Lesson%20Plans/booklets/Eng%20Level%20KG%20Lessons_210816.pdf',
```
mudssrali, line 14: we should avoid having links with `%20` (space)

mudssrali (changes requested):
LGTM - Thanks @ZarishIqbal. I left a question and a suggestion.

web/src/modules/TIP/LessonPlans/List.tsx
ZarishIqbal: You're right. 
I fixed it.

web/src/modules/TIP/TIPConstants.ts
ZarishIqbal: Done.

mudssrali (approved):
Thanks @ZarishIqbal

PR #477, opened 2021-11-22
Optimizations for results, grading, merges etc

hmzahmed:
- Dont send whole questions to db when grading tests ( now sending only if question is correct or not )
- In Quiz Grading, sync only changed values
- In Quiz Grading, send obtained and total marks as a single merge
- Align names to the left when grading quizzes

Note: Decided to change type to for TIPGradedQuestion to Partial of TIPQuesiton because I did not want it to mess with the already existing data


web/src/modules/TIP/FormativeTest/Grading.tsx
```diff
@@ -110,7 +110,6 @@ const Grading: React.FC<PropsType> = ({
 			questions: {
 				...result.questions,
 				[questionId]: {
-					...question,
 					is_correct
 				}
 			}
```
mudssrali, line 115: Are we sure that no question `static-data` isn't being sent to backend?

web/src/actions/index.ts
```diff
@@ -814,24 +814,9 @@ export const saveTIPQuizResult = (
...
-					'targeted_instruction',
-					'quiz_result',
-					grade,
-					subject,
-					quiz_id,
-					'total_marks'
+					quiz_id
 				],
-				value: total_marks
+				value: { obtained_marks, total_marks }
```
mudssrali, line 819: thanks, it looks fine now.

web/src/modules/TIP/FormativeTest/Grading.tsx
```diff
@@ -110,7 +110,6 @@ const Grading: React.FC<PropsType> = ({
 			questions: {
 				...result.questions,
 				[questionId]: {
```
mudssrali, line 112: why we didn't get rid of `...rest.questions`? For existing schools it will keep saving questions static data.

mudssrali (commented):
Thanks @hmzahmed It looks good to me however we need to make sure old existing data doesn't store on our backend again. It's fine to make `Partial<TIPQuestion>` however for existing TIP school we really need to make sure:
- We're not going to store data again
- Everything works fine after this change

Also create a fresh TIP school on local machine and make sure it works fine.

web/src/modules/TIP/FormativeTest/Grading.tsx
hmzahmed: because we still need to save existing questions if user is modifying the quiz marks

hmzahmed:
There was an issue I encountered today because of these changes while working on starting test changes, I have fixed that

mudssrali:
We need to change it entirely that what I was looking forward to.

mudssrali:
> There was an issue I encountered today because of these changes while working on starting test changes, I have fixed that.
> Will test with a local tip school on my machine

So we shouldn't test it to make sure everything works fine? And again, make it priority to test it with existing data and fresh tip school. Make sure we're not going to send question whole static payload to server to store in our db.

mudssrali:
> > There was an issue I encountered today because of these changes while working on starting test changes, I have fixed that.
> > Will test with a local tip school on my machine
> 
> So we shouldn't test it to make sure everything works fine? And again, make it priority to test it with existing data and fresh tip school. Make sure we're not going to send question whole static payload to server to store in our db.

After merging this PR, all questions that has been saved to db, only `is_correct` property would used from them and next-time on saving, would overwrite existing questions static-data. Also all questions IDs that has been saved previous in db must be in CSV otherwise there can be consequences or unwanted bugs would occur.

PR #484, opened 2021-12-07
fixes

ZarishIqbal:
- update quiz icon in results
- upgrade pdfjs worker url version
- align lesson plan print all button


mudssrali (approved):
@ZarishIqbal thanks LGTM

PR #485, opened 2021-12-10
Chore: TIP Patch 2 

hmzahmed:
- Update Status Columns for final test
- Text Tweaks for sorting results

Note: Some bugs were identified as being caused due to id change, and some were fixed by generating new tip json for staging and production, these are not reflected in the code of this PR


mudssrali (approved):
Good work @hmzahmed LGTM 👍🏻

PR #488, opened 2021-12-17
Chore: Fix TIP group sorting bug

ZarishIqbal:
- Pass grade along with `is_correct`


mudssrali (approved):
Thanks @ZarishIqbal  Looks good to me 🎉

PR #496, opened 2021-12-23
Feature: TIP - complete history modal

ZarishIqbal:


mudssrali (changes requested):
Thanks @ZarishIqbal  - Looks good to me. A few comments - We can discuss about timestamp tomorrow sometime.

web/src/modules/TIP/DetailedAnalysis/ClassView/ClassViewCard.tsx
```diff
@@ -184,10 +184,8 @@ const ClassViewCard = ({ std }: ClassViewProps) => {
 								: 'cursor-default text-gray-300'
 						)}
 						onClick={() => {
-							if (std.targeted_instruction?.learning_level) {
-								setModalType(ModalType.STUDENT_HISTORY)
-								setIsComponentVisible(true)
-							}
+							setModalType(ModalType.STUDENT_HISTORY)
+							setIsComponentVisible(std.targeted_instruction?.learning_level && true)
```
mudssrali, line 188: so Instead `string && true` we can use negation operation `!!` to make it boolean.

mudssrali (approved):
Thanks @ZarishIqbal  - Looks good to me just a single change required.

PR #498, opened 2021-12-24
Chore: MIS - Add month filter to monthly-test results

ZarishIqbal:
- added month filter through query params for monthly test.


web/src/pages/exams/results/class/result.tsx
```diff
@@ -474,7 +477,7 @@ class ClassResultsView extends Component<P, S> {
 							scoreSheet={scoreSheet}
 							exams={filtered_exams}
 							smsUrl={url}
-							smsCallback={() => {}}
+							smsCallback={() => { }}
```
mudssrali, line 480: At some point we need to pass handle function for sure.

mudssrali (approved):
Thanks @ZarishIqbal  - This would solve schools' users' issues for entering or viewing marks. LGTM

PR #499, opened 2021-12-27
Chore: TIP - Delete Class Option

ZarishIqbal:
- add delete class check option in Mischool Admin controls


mudssrali (approved):
Thanks @ZarishIqbal  - Looks good to me.

PR #503, opened 2021-12-31
Chore: TIP - bug fixes

ZarishIqbal:
- Fix navigation
- Fix logic for calculating HT dashboard multiple diagnostic test score
- Add spinner to Lesson Plan print all button 
- Design tweaks


web/src/modules/TIP/Classes.tsx
```diff
@@ -12,29 +12,49 @@ type PropsType = P & RouteComponentProps
...
+			return `/${url[1]}/${url[2]}/${sectionId}/${class_name}`
+		} else if (url[2] === 'oral-test') {
+			return `/${url[1]}/${url[2]}`
+		} else if (url[2] === 'quizzes') {
+			return `/${url[1]}/${url[2]}/${class_name}`
+		} else if (url[2] === 'quiz-result') {
+			return `/${url[1]}/${url[2]}/${class_name}`
+		} else {
+			return `/${url[1]}/${url[2]}/${class_name}`
+		}
```
mudssrali, line 38: This pattern is observed many-times. The general rule of thumb is we always avoid breaking circuits e.g. `switch-case`, `if else if...` Here in this case, we're returning using `return` and this equivalent to
```tsx
if (url[2] === 'diagnostic-result') {
    return `/${url[1]}/${url[2]}/${sectionId}/${class_name}`
} 
if (url[2] === 'formative-result' || url[2] === 'summative-result') {
     return `/${url[1]}/${url[2]}/${class_name}`
} 
if (url[2] === 'lesson-plans') {
     return `/${url[1]}/${url[2]}/${class_name}`
} 
if (url[2] === 'diagnostic-test') {
     return `/${url[1]}/${url[2]}/${sectionId}/${class_name}`
}
if (url[2] === 'oral-test') {
     return `/${url[1]}/${url[2]}`
} 
if (url[2] === 'quizzes') {
    return `/${url[1]}/${url[2]}/${class_name}`
} 
if (url[2] === 'quiz-result') {
    return `/${url[1]}/${url[2]}/${class_name}`
}

return `/${url[1]}/${url[2]}/${class_name}`
```
Also we change into object-mapping where each test would be key and path would be value as follow
```tsx
const routes = {
'diagnostic-result': `/${url[1]}/${url[2]}/${sectionId}/${class_name}`,
'formative-result': `/${url[1]}/${url[2]}/${class_name}`,
'summative-result': `/${url[1]}/${url[2]}/${class_name}`,
'diagnostic-test': `/${url[1]}/${url[2]}/${sectionId}/${class_name}`,
'oral-test': `/${url[1]}/${url[2]}`,
'quizzes': `/${url[1]}/${url[2]}/${class_name}`,
'quiz-result': `/${url[1]}/${url[2]}/${class_name}`
} 

return routes[url[2]] ?? `/${url[1]}/${url[2]}/${class_name}`
```
Again in `object-mapping` solution we can see there's repetition for routes. The best solution could be `switch-case` as follow 
```tsx
let route = null
switch(url[2]) {
  // Multi-case: single operation
  case: 'diagnostic-test':
  case: 'diagnostic-result':
         route = `/${url[1]}/${url[2]}/${sectionId}/${class_name}`
         break
   default:
         route = `/${url[1]}/${url[2]}/${class_name}`
}

return route
```
To read more about: [Switch-Case](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch)

web/src/modules/TIP/DiagnosticTest/Result/index.tsx
```diff
@@ -16,9 +16,8 @@ interface P {
 type PropsType = P & RouteComponentProps
 
 const DiagnosticTestResult: React.FC<PropsType> = ({ match, targeted_instruction, classes }) => {
-	const [sectionId, setSectionId] = useState('')
 	const url = match.url.split('/')
-
+	const [sectionId, setSectionId] = useState(url[3] ?? '')
```
mudssrali, line 20: By looking at `url[3]` - we cannot tell what will we get once it executed. Aren't we're using match params e.g. school/:id for these type of URLs?

web/src/modules/TIP/DiagnosticTest/Result/index.tsx
```diff
@@ -39,6 +38,7 @@ const DiagnosticTestResult: React.FC<PropsType> = ({ match, targeted_instruction
 			return valid_classes[s.className.toLowerCase()]
 		})
 
+	useEffect(() => setSectionId(url[3] ?? ''), [url])
```
mudssrali, line 41: why we're invoking again `setSectionId` inside useEffect(...) ?

web/src/modules/TIP/DiagnosticTest/index.tsx
```diff
@@ -17,9 +17,10 @@ type PropsType = P & RouteComponentProps
 
 // We can only do diagnostic tests for Classes which match the string they are assigned to
 const DiagnosticTest: React.FC<PropsType> = ({ match, targeted_instruction, classes }) => {
-	const [sectionId, setSectionId] = useState('')
 	const url = match.url.split('/')
+	const [sectionId, setSectionId] = useState(url[3] ?? '')
 
+	useEffect(() => setSectionId(url[3] ?? ''), [url])
 	const valid_classes = Object.values(targeted_instruction.tests)
```
mudssrali, line 24: same goes here.

web/src/modules/TIP/FormativeTest/Result/index.tsx
```diff
@@ -13,9 +13,10 @@ interface P {
 type PropsType = P & RouteComponentProps
 
 const FormativeTestResult: React.FC<PropsType> = ({ match, targeted_instruction }) => {
-	const [class_name, setClassName] = useState('')
 	const url = match.url.split('/')
+	const [class_name, setClassName] = useState(url[3] ?? '')
 
+	useEffect(() => setClassName(url[3] ?? ''), [url])
```
mudssrali, line 19: here again

web/src/modules/TIP/HTDashboard/Tests/DiagnosticTest/index.tsx
```diff
@@ -53,10 +53,12 @@ const HTDiagnostic = () => {
...
 	const breakdown = Object.values(filteredStudents).reduce((resultBreakdown, curr) => {
+		const subjectTestTaken: string[] = []
 		for (const [resultId, result] of Object.entries(curr.targeted_instruction?.results ?? {})) {
-			if (result.checked && result.type === view) {
-				resultBreakdown[targeted_instruction.tests[resultId]?.subject] =
-					(resultBreakdown[targeted_instruction.tests[resultId]?.subject] || 0) + 1
+			const subject = targeted_instruction.tests[resultId]?.subject
+			if (result.checked && result.type === view && !subjectTestTaken.includes(subject)) {
+				subjectTestTaken.push(subject)
+				resultBreakdown[subject] = (resultBreakdown[subject] ?? 0) + 1
```
mudssrali, line 61: We could use `reduce(...)` inside `reduce(...)` as well.

web/src/modules/TIP/HTDashboard/Tests/DiagnosticTest/index.tsx
```diff
@@ -67,12 +69,15 @@ const HTDiagnostic = () => {
 		if (!isValidStudent(curr)) {
 			return bd
 		}
-		const t = Object.entries(curr.targeted_instruction?.results || {})
+		const subjectTestTaken: string[] = []
+		Object.entries(curr.targeted_instruction?.results || {})
```
mudssrali, line 73: Instead of `||`, it must be `??`

web/src/modules/TIP/LessonPlans/index.tsx
```diff
@@ -13,9 +13,10 @@ interface P {
 type PropsType = P & RouteComponentProps
 
 const LessonPlans: React.FC<PropsType> = ({ match, targeted_instruction }) => {
-	const [class_name, setClassName] = useState('')
 	const url = match.url.split('/')
+	const [class_name, setClassName] = useState(url[3] ?? '')
 
+	useEffect(() => setClassName(url[3] ?? ''), [url])
```
mudssrali, line 19: Here goes again.

web/src/modules/TIP/Classes.tsx
```diff
@@ -12,29 +12,49 @@ type PropsType = P & RouteComponentProps
...
 // for formative, grades will be tiplevels
 // for diagnostic, grades will be tipgrades
-const Classes: React.FC<PropsType> = ({ setSectionId, sortedSections, match }) => {
+const Classes: React.FC<PropsType> = ({ sortedSections, match }) => {
 	const url = match.url.split('/')
 
 	// "grade" is misleading as we only deal with TIPLevels here. we map the level to a
 	// TIPGrade inside the map function below (mapped_grade)
+	const getRoute = (url: string[], sectionId: string) => {
+		const class_name = getClassnameFromSectionId(sortedSections, sectionId)
```
mudssrali, line 21: Instead of `class_name`, we cab use `className`

mudssrali (changes requested):
Thanks @ZarishIqbal  - I see some good improvements for navigation. We need to test it well before pushing it master. I left a few suggestions, please check them.

web/src/modules/TIP/DiagnosticTest/Result/index.tsx
ZarishIqbal: What's happening here is that child component amends `url` and upon that change I want the `sectionId` to be updated as well.

web/src/modules/TIP/HTDashboard/Tests/DiagnosticTest/index.tsx
ZarishIqbal: This is existing logic that I didn't change. The main goal was to count just 1 test for each subject for each student. Can we optimize that logic in any way using reduce?

PR #509, opened 2022-01-11
Design Fix: Range Slider for Quiz Grading

hmzahmed:
This PR contains a design fix for the range slider, it works in the following way

1. Makes the range input background transparent
2. Make its width a little smaller than full, make its position absolute and place it in the center
3. Create another div whose width will be controlled by obtained marks
4. place that div under the range input so when the width changes, it looks like progress of range slider


web/src/modules/TIP/Quizzes/SingleStdGrading.tsx
```diff
@@ -28,22 +28,34 @@ const SingleStdGrading: React.FC<P> = ({
...
+	// this hook fires before DOM has been painted
+	useLayoutEffect(() => {
+		// Custom logic by trail and error to find the perfect values for percentage of width
+		const progressBar = document.getElementById(`${student.id}marks`)
+
+		// We have to show the whole input as red in case of absent which is -1, hence simply 100 is returned
+		// In case of other inputs we use (obtainedMarks + (obtainedMarks > 4 ? 0 : 1))
+		progressBar.style.width = `${
+			obtainedMarks === -1 ? 100 : (obtainedMarks + (obtainedMarks > 4 ? 0 : 1)) * 10
+		}%`
```
mudssrali, line 56: Instead of doing `string-interpolation`, we can convert integer to string with `toString()`

web/src/modules/TIP/Quizzes/SingleStdGrading.tsx
```diff
@@ -28,22 +28,34 @@ const SingleStdGrading: React.FC<P> = ({
...
+			'10': 28
 		}
 
 		return conversionMap[marks]
 	}
 
+	// this hook fires before DOM has been painted
+	useLayoutEffect(() => {
+		// Custom logic by trail and error to find the perfect values for percentage of width
+		const progressBar = document.getElementById(`${student.id}marks`)
```
mudssrali, line 50: nice :)

mudssrali, line 50: instead of apply properties like this, we could use `useRef()`.

mudssrali (changes requested):
Thanks @hmzahmed  - It looks good to me. Interesting use of `useLayoutEffect`. Just two comments.

mudssrali (approved):
Looks good to me. Thanks @hmzahmed

PR #510, opened 2022-01-12
Bug Fix: Quiz Library Checkmark

hmzahmed:
The bug was that checkmarks were showing up on quizzes where there were no students, updated checkQuizMarked method to check for students array length as well because ".every( )" returns true in case of empty arrays


web/src/utils/TIP.ts
```diff
@@ -1034,11 +1034,16 @@ export const checkQuizMarked = (
...
+		students.every(student => {
+			const obtainedMarks =
+				student.targeted_instruction?.quiz_result?.[learningLevel]?.[subject]?.[quizId]
+					?.obtained_marks ?? ''
 
-		return `${obtainedMarks}` !== ''
-	})
+			return `${obtainedMarks}` !== ''
+		})
+	)
```
mudssrali, line 1048: it's better to return first
```tsx
if (students.length === 0) {
    return false
}
```

mudssrali (changes requested):

web/src/utils/TIP.ts
hmzahmed: It does not matter since if left hand side of && is false, it will not even evaluate the right hand side and return false immediately

mudssrali:
@hmzahmed For the sake of readability, it's better to return earlier

PR #513, opened 2022-01-17
Chore: Tip - Show unsynced changes

ZarishIqbal:
- Add catch to fetchTargetedInstruction
- Add sync icon and unsynced changed count to TIP Header


web/src/modules/TIP/Header.tsx
```diff
@@ -1,16 +1,23 @@
...
+import React, { useState } from 'react'
+import { useSelector } from 'react-redux'
 import { Link, RouteComponentProps, withRouter } from 'react-router-dom'
 
 type PropsType = RouteComponentProps
 
 const Header: React.FC<PropsType> = ({ history }) => {
+	const queuedChanges = Object.values(
+		useSelector((state: RootReducerState) => state.queued.mutations)
+	)
```
mudssrali, line 15: we can do this as well, which is more intuitive
```tsx
const queuedChanges = useSelector((state) => Object.key(state.queued.mutations).length)
```

web/src/modules/TIP/Header.tsx
```diff
@@ -23,12 +30,63 @@ const Header: React.FC<PropsType> = ({ history }) => {
 									onClick={() => history.goBack()}>
 									<ArrowLeftIcon className="w-6" />
 								</div>
+								<div className="flex space-x-3 items-center justify-center">
+									<Link to={'/targeted-instruction'} className=" flex ">
+										<img className="h-8" src={TIP} />
```
mudssrali, line 35: missing `alt` attribute

web/src/modules/TIP/Header.tsx
```diff
@@ -23,12 +30,63 @@ const Header: React.FC<PropsType> = ({ history }) => {
...
+														<div className="flex justify-center items-center h-full w-full space-x-3">
+															<Spinner />
+															<span>Reconnecting</span>
+														</div>
+													) : (
+														<>
+															<RefreshIcon className="h-6 w-2/12" />
+															<div className="flex flex-col w-10/12">
+																<h1 className=" font-bold">
+																	You are not online
```
mudssrali, line 59:
```txt
You are offline
```
OR
```txt
You are not online
```

web/src/modules/TIP/Header.tsx
```diff
@@ -23,12 +30,63 @@ const Header: React.FC<PropsType> = ({ history }) => {
...
+																<button
+																	className="flex mt-2 text-light-blue-tip-brand underline items-center"
+																	onClick={() => {
+																		setReconnecting(true)
+																		setTimeout(
+																			() =>
+																				window.location.reload(),
+																			1000
+																		)
+																	}}>
```
mudssrali, line 76: are we sure about `timeout: 1000` ? If yes, We don't need to have `setReconnecting(true)` bc it would create bad UX.

mudssrali (changes requested):
Thanks @ZarishIqbal  - Nice work. Just a few suggestion and a single change.

web/src/modules/TIP/Header.tsx
ZarishIqbal: The design has a reconnecting spinner. Timeout is to show that spinner and then reload the page.

mudssrali (approved):
Thanks @ZarishIqbal

hmzahmed (approved):

PR #516, opened 2022-01-21
Chore: MIS Fixes

hmzahmed:
Bug fixes for Diary Module

- Include subjects in diary only if they are also in class
- increase margin of diary tables when printing


web/src/pages/school/diary/index.tsx
```diff
@@ -67,12 +67,25 @@ const Diary: React.FC = () => {
 
 	const currentDate = moment(state.selectedDate, 'YYYY-MM-DD').format('DD-MM-YYYY')
 
+	const getFilteredDiary = (diary: {
+		[subject: string]: {
+			homework: string
+		}
+	}) => {
+		return Object.entries(diary).reduce((agg, [subject, diary]) => {
+			if (!Object.keys(classes?.[state.classId].subjects).includes(subject)) {
```
mudssrali, line 76: we have to put guard here `Object.keys(classes?.[state.classId]?.subjects ?? {})`

web/src/pages/school/diary/index.tsx
```diff
@@ -67,12 +67,25 @@ const Diary: React.FC = () => {
 
 	const currentDate = moment(state.selectedDate, 'YYYY-MM-DD').format('DD-MM-YYYY')
 
+	const getFilteredDiary = (diary: {
+		[subject: string]: {
+			homework: string
+		}
+	}) => {
```
mudssrali, line 74: We might have typings for `diary` somewhere in the file or `typings.d.ts`. If we don't have, create it separate outside the component and then apply here.

mudssrali (changes requested):
Thanks @hmzahmed  - Just two quick changes required

PR #518, opened 2022-01-27
Chore: TIP-HT Dashboard Changes 

salehcerp:
### Fixed 

- changed levels to progress
- show null instead of NaN in HT Dashboard


web/src/modules/TIP/HTDashboard/Tests/SummativeTest/index.tsx
```diff
@@ -225,64 +228,34 @@ const HTSummative = () => {
...
+									<div className="text-xs text-gray-900 place-self-start self-center pl-2">
+										{gradeToColorMapper[k]} Group
+									</div>
+									{TIPSubjects.map(subject => {
+										const percentage = getPercentageFromMarks(
+											item?.[subject]?.obt,
+											item?.[subject]?.total
+										)
+										return (
+											<div className="">
```
hmzahmed, line 245: What purpose is this div serving?

web/src/modules/TIP/HTDashboard/Tests/SummativeTest/index.tsx
```diff
@@ -225,64 +228,34 @@ const HTSummative = () => {
...
+												<button
+													className={clsx(
+														'text-xs text-center shadow-md flex justify-center items-center rounded-md  w-12 h-10',
+														getColorClassFromPercent(percentage)
+													)}>
+													{!isNaN(percentage) ? percentage + '%' : ''}
+												</button>
```
hmzahmed, line 252: This UI element is not doing any action nor does it need any click functionality, we can change this from button to div

web/src/modules/TIP/HTDashboard/Tests/SummativeTest/index.tsx
```diff
@@ -225,64 +228,34 @@ const HTSummative = () => {
...
-													item?.['Maths']?.total
-												)
-											)
-										)}>
-										{getPercentageFromMarks(
-											item?.['Maths']?.obt,
-											item?.['Maths']?.total
-										) + '%'}
-									</button>
+						{Object.entries(items).map(([k, item], index) => {
```
hmzahmed, line 231: you can rename the variable 'k' to 'grade', we should always name variables something descriptive, 'item' can also be renamed to 'result'

web/src/modules/TIP/HTDashboard/Tests/SummativeTest/index.tsx
```diff
@@ -225,64 +228,34 @@ const HTSummative = () => {
...
+							return (
+								<div
+									key={index}
+									className="grid grid-cols-4 my-2 justify-items-center">
```
hmzahmed, line 235: We should avoid using index as key if we can, here we can use grade as index

web/src/modules/TIP/HTDashboard/Tests/SummativeTest/index.tsx
hmzahmed, line 252: We should also provide a key whenever we render a list of elements, here we can provide a key by concatenating grade and subject

hmzahmed (changes requested):
Congrats on your first PR, Nice work!! everything looks good. Just a few suggestions

mudssrali (approved):
Looks good to me. Thanks @salehcerp

PR #524, opened 2022-02-21
Chore: TIP - Add staff ids

ZarishIqbal:
### This PR contains
- [x] add `tipUniqueId` in staff type
- [x] create hook `useTIPAccess` for TIP Access
- [x] add unique Id field to teacher profile


web/src/pages/staff/create.tsx
```diff
@@ -97,7 +96,18 @@ export const CreateOrUpdateStaff = () => {
 		if (!isValidPhone(profile.Phone) && profile.Phone !== '') {
 			return toast.error('Please provide correct phone number!')
 		}
+		if (profile.tipUniqueId) {
+			let errorMessage = 'This TIP Id is already assigned to an existing teacher'
+
+			for (const teacher of Object.values(staff ?? {})) {
+				const isDuplicateId =
+					teacher.tipUniqueId === state.profile.tipUniqueId && profile.id !== teacher.id
```
mudssrali, line 104: why we're anding here?
```
&& profile.id !== teacher.id
```

web/src/pages/staff/create.tsx
```diff
@@ -97,7 +96,18 @@ export const CreateOrUpdateStaff = () => {
...
 
+				if (isDuplicateId) {
+					return toast.error(errorMessage)
+				}
```
mudssrali, line 108: instead of returning from for..of, would be best if you break the loop and check outside and return

mudssrali (changes requested):
Thanks looks good to me, Thanks @ZarishIqbal however a change and a question in the comments

web/src/pages/staff/create.tsx
ZarishIqbal: To remove the teacher we are editing at the moment?

PR #529, opened 2022-02-26
Chore: Update SLOs algorithm

ZarishIqbal:
### This PR contains the following changes:

- [x] update slos for results 
> - For `diagnostic` tests, we convert the classes into grades and check the questions according to `questionGrade <= classGrade`
> - For `summative` tests, we consider the groups and check the questions according to `questionGrade === group`
- [x] update slos for HT dashboard 
> - we consider the question against the selected `grade` however, we also consider `oral` tests for blue group
> - We preload all the slos before calculating the percentages


web/src/modules/TIP/HTDashboard/Tests/ComparisonMatrix/skill-view.tsx
```diff
@@ -39,39 +39,37 @@ export const MatrixSkillView = ({ filters }: SkillViewProps) => {
...
+				})
 
 				// When calculating result, only those questions are chosen whose grade is equal to
 				// or lower than the grade of the student whose result is being calculated
+
 				if (res.checked) {
 					Object.entries(res.questions).forEach(([questionId, question]) => {
 						const test = tests?.[id]
-						const { type, questions } = test
+						const { questions, type } = test
```
mudssrali, line 53: while object destructuring, order doesn't matter.

mudssrali (approved):
Looks fine to me @ZarishIqbal  - Thanks

PR #531, opened 2022-03-02
Chore: Web Cleanup

mudssrali:
This PR introduces huge changes to `mis/web`, however in more broader sense following changes have been made:

- [x] Clean up `modules` and move stuff to `pages` dir
- [x] Rename all files to `kebab-case` to create standard to follow while creating new files
- [x] Remove unused and consolidate dependencies
- [x] Code Refactoring (from changing names to writing entire files)
- [x] Run prettier (auto+manual) and fix most of the warning  

**Why we did this**?

When this project started, we added most of code under `modules` directory. After 2 years, when we started redesigning the whole MISchool PWA, meanwhile we had to work on *Targeted Instruction Program* tool withing the MISchool alongside redesigning. To manage code migrations  under single `pages` directory we have to hold on for initial releases and interventions.

To make sure new developers can be easily  onboard-ed without any issue, we decided to consolidate inconsistencies among coding standards, directories/files naming, variable naming convention and a lot more. A few general rules has been followed:

- File or Directory name must be lowered `kebab-case`
- Component would be `pascal-case`
- Function and variables must be `camel-case`
- Functional component or simple function must be `named export` (we still have default export in a few files)
- Reusable constants must be put under `web/src/constants`
- To access something from store, We've created a few hooks under `web/src/hooks` (worth mentioning, for hooks naming convention, we followed react approach e.g. useStudents, useClasses
- We also make sure our code base always upgrade when there major release of any dependency available


web/src/constants/targeted-instruction/index.ts
```diff
@@ -48,3 +48,24 @@ export enum PeerViewType {
 	RESULT_VIEW = 'RESULT_VIEW',
 	PEER_GROUP_VIEW = 'PEER_GROUP_VIEW'
 }
+
+//Need gradeMap to assign numerical values to grades, this is done so that we can filter out questions
+//that are of higher grade than the students current grade when calculating result
+export const gradeMap = {
+	KG: 0,
```
ZarishIqbal, line 55: This  needs to be `KG: 1`.

web/src/actions/index.ts
```diff
@@ -369,7 +366,7 @@ export const createSchoolLogin =
 		syncr
 			.send({
 				type: 'LOGIN',
-				client_type,
+				clientType,
```
ZarishIqbal, line 369: I couldn't login so had to change this to  `client-type: clientType`

ZarishIqbal (changes requested):
Tested TIP modules.
Left comments on the issues I found.

web/src/actions/index.ts
mudssrali: Good catch 💚

web/src/constants/targeted-instruction/index.ts
mudssrali: Or we just need to do KG:1 ?

ZarishIqbal: Just KG: 1. This is because for KG .. we consider questions of grade 1 in diagnostics tests..

PR #532, opened 2022-03-02
Chore: TIP - Single Student Progress

ZarishIqbal:
### This PR contains:

- A single student view to get the progress of a student in each SLO. 
- The subject and group filters do apply to these SLOs in results.
- The navigation added is through route parameters instead of components, to add the option of going back to class view.


web/src/pages/targeted-instruction/head-teacher-dashboard/tests/comparison-matrix/index.tsx
```diff
@@ -1,5 +1,6 @@
 import Tabs from 'components/tabs'
 import React, { useState } from 'react'
+import { useHistory, useRouteMatch } from 'react-router'
```
mudssrali: we've removed this. We must be importing from `react-router-dom` next time

mudssrali (approved):
Looks good to me @ZarishIqbal Thanks.

PR #538, opened 2022-03-15
Chore: TIP - Patch 8

ZarishIqbal:
### This PR contains the following changes:
- [x] Single Student Performance in the Comparison matrix - Ungraded SLOs are greyed out without a percentage
- [x] Student Learning Progress - Remedial-wise percentage tabs under each subject are clickable and take users to the `Skill View page` in Summative Results


web/src/pages/targeted-instruction/head-teacher-dashboard/tests/summative-test/index.tsx
```diff
@@ -220,7 +221,14 @@ const HTSummative = () => {
 										gradeResult?.[subject]?.obt,
 										gradeResult?.[subject]?.total
 									)
-									return <PercentageBox key={subject} percentage={percentage} />
+
+									return (
+										<Link
+											key={subject + level}
+											to={`/targeted-instruction/summative-result/Level%20${level}/${subject}/result`}>
```
mudssrali, line 228: why we're using  `%2` ? adding space here?

mudssrali (approved):
Thanks @ZarishIqbal  Looks good to me.

web/src/pages/targeted-instruction/head-teacher-dashboard/tests/summative-test/index.tsx
ZarishIqbal: It's actually %20.. level mein we're getting Kg, 1, 2, 3. 

We're supposed to pass Level KG, Level 1 and so on. 

%20 is the space we need.

mudssrali: at some point, we should get rid of these spaces and sanitize url

PR #539, opened 2022-03-28
Chore: MIS - Bug Fixes

ZarishIqbal:
### This PR fixes the following:
- [x] Simplify grade calculation 
- [x] Set `year` as a number in expenses
- [x] Render exams are based on student exams instead of class exams.
- [x] Instead of passing `sectionId` and `sectionTeacher`, extract this information within the print card component.  
- [x] Update mobile view links in analytics.


web/src/pages/exams/results/student/result.tsx
```diff
@@ -67,8 +75,7 @@ const SingleStudentResult = () => {
 
 			if (
 				examYear === Number(year) &&
-				(exam.name === ExamTerms.TEST || exam.name === ExamTerms.MONTHLY ? false : true) &&
-				exam.section_id === state.section
+				(exam.name === ExamTerms.TEST || exam.name === ExamTerms.MONTHLY ? false : true)
```
mudssrali, line 78: We could  write only
```tsx
if (exam.name === ExamTerms.TEST || exam.name === ExamTerms.MONTHLY)
```

web/src/pages/exams/results/student/result.tsx
```diff
@@ -37,16 +35,26 @@ type AugmentedGroupedExams = {
...
-	const { students, exams, faculty, sms_templates, settings, classes, assets } = useSelector(
-		(state: RootReducerState) => state.db
-	)
+	const {
+		students,
+		exams: misExams,
+		sms_templates,
+		settings,
+		assets
+	} = useSelector((state: RootReducerState) => state.db)
```
mudssrali, line 44: First we've AppSelector and we will be using it always. Instead of single hook, we should split into multiple as
```
const students = useStudents()
.
.
.
```

web/src/pages/exams/results/student/result.tsx
```diff
@@ -120,21 +125,18 @@ const SingleStudentResult = () => {
...
 		if (
 			exam.name ===
-				(state.type === ExamTypes.MONTHLY
-					? state.examName?.split(' ')[1]
-					: state.examName) &&
-			moment(exam.date).format('YYYY') === state.year.toString() &&
+			(state.type === ExamTypes.MONTHLY
+				? state.examName?.split(' ')[1]
+				: state.examName) &&
+			moment(exam.date).format('YYYY') === state.year?.toString() &&
```
mudssrali, line 134:
```tsx
moment(exam.date).year() === state.year
```

web/src/pages/expense/index.tsx
```diff
@@ -313,13 +313,12 @@ export const Expense = () => {
...
-																			expense.date
-																		).format('ddd')} ${moment(
-																			expense.date
-																		).format('Do')}`}</p>
+																		<p>{`${expense.label
+																			} - ${moment(
+																				expense.date
+																			).format('ddd')} ${moment(
+																				expense.date
+																			).format('Do')}`}</p>
```
mudssrali, line 321: We could move it before return

web/src/pages/home/statistics.tsx
```diff
@@ -29,9 +29,7 @@ type PropTypes = {
 }
 
 export const getDailyStatsRoute = (stats_type: string) => {
-	return `/analytics/${stats_type}?start_date=${moment().format(
-		'MM-DD-YYYY'
-	)}&end_date=${moment().format('MM-DD-YYYY')}&period=Daily`
+	return `/analytics/mobile`
```
mudssrali, line 32: why we did this? I remember that I did for a reason

mudssrali (changes requested):
Thanks @ZarishIqbal - Just a minor changes.

web/src/pages/exams/results/student/result.tsx
ZarishIqbal: state.year is a number.

web/src/pages/home/statistics.tsx
ZarishIqbal: I believe either we used these query params in old analytics or you wanted to incorporate them is the new one but we never got to that.

PR #546, opened 2022-04-20
Chore - TIP Patch 10

ZarishIqbal:
This PR contains fixes for the following issues:

### Oral Test Students list 
The following use cases need to be handled

1. Student was sorted into `oral` group. The student needs to appear ungraded for all the versions.
2. Student was sorted into `oral` group and marked i.e the current grade isn't `oral test`. For the marked version, the student should appear graded. For the rest of the versions, the student needs to appear ungraded.
3. Student was sorted into `oral` group and sorted again into another group. The student shouldn't appear on the oral group grading page.

Solution:

- [x] Checking the current grade handles case 1.
- [x] Checking the sorting history for `Oral Test` handles case 2.
- [x] Checking the result eliminates the students in case 3 who had a history of being sorted in `Oral Test` but weren't graded.

### Remove duplicate students from student csv when importing students.
For each student in the CSV `Name, RollNumber, BForm, Phone, Gender, ManCNIC, ManName` are checked against the existing students. If a student with similar data already exists, the student is filtered out from the CSV.


web/src/constants/targeted-instruction/index.ts
```diff
@@ -91,3 +91,8 @@ export enum TIPTeacherAnalyticEvents {
 	PEER_GROUP = 'PEER_GROUP',
 	SUMMATIVE_SCORE = 'SUMMATIVE_SCORE'
 }
+export const TIPSubjectMap: { [subject: string]: string } = {
```
mudssrali, line 94: More elegant way to do:
```tsx
export const TIPSubjectMap: Record<string, string> = { }
```

web/src/pages/students/add/excel-import.tsx
```diff
@@ -67,65 +68,101 @@ export const ImportStudentsCSV = () => {
...
+									s.Phone === formatPhone(Phone) &&
+									s.Gender ==
+									(Gender.toLowerCase() === 'm'
+										? 'male'
+										: Gender.toLowerCase() === 'f'
+											? 'female'
+											: '') &&
+									s.ManCNIC === formatCNIC(ManCNIC) &&
+									s.ManName === toTitleCase(ManName)
+							)
```
mudssrali, line 102: Instead of checking for existing student like this, I would say just check Roll-Number and Admission-Number.

web/src/pages/students/add/excel-import.tsx
```diff
@@ -67,65 +68,101 @@ export const ImportStudentsCSV = () => {
...
-						return student
-					}
-				)
+							return student
+						}
+					)
 
+				if (items.length > students.length) {
+					toast.error('Duplicate student were removed!')
+				}
```
mudssrali, line 165:
```tsx
toast.error('Duplicate students were removed automatically!'
```

web/src/utils/tip.ts
```diff
@@ -33,9 +33,27 @@ export const getOralTestStudents = (
...
+		//
+		//	2-	Student was sorted into Oral group and marked i.e the current grade isn't oral test.
+		//		For the marked version the student should appear graded. For the rest of the versions, student needs to appear as ungraded
+		//
+		//	3-	Student was sorted into Oral group and sorted again into another group.
+		//		The student shouldn't appear in the oral group grading page
+		//
+		// Checking the current grade handles case 1.
+		// Checking the sorting history handles case 2.
+		// Checking the result eliminated the students in case 3.
```
mudssrali, line 48: Nice 💚 - Little indentation required.

mudssrali (changes requested):
Thanks @ZarishIqbal  - A few quick changes.

web/src/pages/students/add/excel-import.tsx
ZarishIqbal: @mudssrali Should I do something like this then? Seems more compact!
```js
.filter(([Name, RollNumber, ...rest]) =>
    !MISStudents.find(s =>
        s.Name === toTitleCase(Name) &&
	s.RollNumber == RollNumber &&
	s.AdmissionNumber == rest[rest.length - 1]
	)
)
```

mudssrali: Yes - also no need to have Name check as well. There may be case where actually they change name and rollnumber and admission number would be same - this won't work in that case. so it's safe to go with roll and admission number conditions

PR #551, opened 2022-05-24
Create and delete class button refactors

salehcerp:
The following changes were made:

- used spinner instead of loader in class delete button
- refactored visibility of delete class button


hmzahmed (approved):
Looks good to me!

PR #552, opened 2022-05-27
TIP  JSON Download Notification

mudssrali:


web/src/pages/targeted-instruction/index.tsx
```diff
@@ -31,26 +36,35 @@ enum Layouts {
...
 			dispatch(fetchTargetedInstruction())
 		}
-	}, [])
+	}, [hasTIPAccess])
+
+	const handleClickBtnDownload = () => {
+		dispatch(fetchTargetedInstruction())
+	}
+
+	const tipDataNotFetched = Object.keys(targetedInstruction).length === 0
```
ZarishIqbal, line 64: Instead of checking the length of keys for `targetedInstruction`, we need to check the length for an inner object like `curriculum`. 
Reason: we assign default empty objects to the inner variables of `targetedInstruction`. 

`const tipDataNotFetched = Object.keys(targetedInstruction?.curriculum).length === 0`

ZarishIqbal (changes requested):
Just one change is required.

web/src/pages/targeted-instruction/index.tsx
mudssrali: It's fine if we check keys only reason is that either we get the complete JSON or not. If we successfully get the JSON data then we store it otherwise it would be empty object.

web/src/actions/index.ts
```diff
@@ -1,6 +1,12 @@
 import Dynamic from '@cerp/dynamic'
 import Syncr from '@cerp/syncr'
-import { ActionTypes, clientType, OnboardingStage, RootDBMergePrefix } from 'constants/index'
+import {
+	ActionTypes,
+	clientType,
```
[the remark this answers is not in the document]
mudssrali: Need to check.

PR #555, opened 2022-06-05
Improve Delete Class with Sections

mudssrali:
This PR show improvements and bugfix for #553


web/src/actions/index.ts
```diff
@@ -463,12 +463,23 @@ export const deleteClass =
...
+				...deletes,
+				{ path: ['db', 'settings', 'classes', 'additionalFees', misClass.id] },
+				{ path: ['db', 'settings', 'classes', 'defaultFee', misClass.id] },
+				{ path: ['db', 'classes', misClass.id] }
+			]
+		} else {
+			deletes = [
+				...deletes,
+				...Object.keys(misClass.sections).map(secId => ({
+					path: ['db', 'classes', misClass.id, 'sections', secId]
```
ZarishIqbal, line 479: You should save class information in class history before deleting it.

web/src/pages/classes/create.tsx
```diff
@@ -530,17 +533,9 @@ const DeleteClassModal = ({
...
-	const totalStudents = useMemo(
-		() => getStudentsInClass(students, sectionIds).length,
-		[students, sectionIds]
-	)
-
 	const alertText = tipUser
-		? ''
-		: 'All fee settings for this class will also be gone. Are you sure you want to proceed?'
+		? 'Are you sure you want to proceed?'
+		: 'All fee settings for this class will also be removed. Are you sure you want to proceed?'
```
ZarishIqbal, line 538: Fee settings aren't removed in the case when all sections are not deleted.

ZarishIqbal (commented):
I have made two suggestions. Kindly, check them out.

web/src/actions/index.ts
mudssrali: We would implement this at some point.

web/src/pages/classes/create.tsx
mudssrali: I intentionally left this. We will take this later.

PR #558, opened 2022-06-08
Remove Test Password

mudssrali:
This is initial thought for password removal and handling test ids.


web/src/pages/targeted-instruction/formative-test/insert-grades.tsx
```diff
@@ -16,56 +18,77 @@ import {
...
+import toTitleCase from 'utils/to-title-case'
 
 const InsertGrades = () => {
 	const students = useStudents()
 	const match = useRouteMatch()
 	const url = match.url.split('/')
+	const { class_name: stdClassName, subject, section_id, test_id } = match.params as Params
 
 	const [state, setState] = useState({ searchText: '' })
+	const [testId, setTestId] = useState<string>(test_id)
```
ZarishIqbal, line 30: The default value can be `testIds[0]` so that it is not randomly selected.

web/src/hooks/useTIPTestIds.ts
```diff
@@ -0,0 +1,24 @@
+import { useAppSelector } from './useAppSelector'
+
+export const useTIPTestIds = (stdClassName: string, subject: string, testType: string) => {
+	const tests = useAppSelector(state => state.targeted_instruction.tests)
```
ZarishIqbal, line 4: We need to handle the case for oral test. The test type in csv is `diag`nostic but the testType passed is `oral-test`

ZarishIqbal (approved):
The initial flow for the password modal removal seems to be working fine.
:+1: 

I do have a few suggestions and a question. 
Do we know which result is being reflected on the results page for the students for whom we have entered scores across multiple versions?

